Issue #8, opened 2022-02-13
Would it be possible to apply the Rich formatting without monkey patching


ewels:
Would it be possible to apply the Rich formatting without monkey patching?

I was thinking that the `rich_click` namespace could replicate the `click` namespace. But supply customized version of Click objects (via inheritance).

This is what we have currently:

```python
import click
import rich_click

class RichClickGroup(click.Group):
    def format_help(self, ctx, formatter):
        rich_click.rich_format_help(self, ctx, formatter)
class RichClickCommand(click.Command):
    def format_help(self, ctx, formatter):
        rich_click.rich_format_help(self, ctx, formatter)

@click.group(cls=RichClickGroup)
@click.option('--debug/--no-debug', default=False)
def cli(debug):
    click.echo(f"Debug mode is {'on' if debug else 'off'}")

@cli.command(cls=RichClickCommand)
def sync():
    click.echo('Syncing')
```

But I think it could be as simple (for the user at least) as this:

```python
import rich_click as click 

@click.group()
@click.option('--debug/--no-debug', default=False)
def cli(debug):
    click.echo(f"Debug mode is {'on' if debug else 'off'}")

@cli.command()
def sync():
    click.echo('Syncing')
```

Thoughts?

_Originally posted by @willmcgugan in https://github.com/ewels/rich-click/discussions/7_

ewels:
Done in #10

PR #10, opened 2022-02-13
Import as

ewels:
Start playing with the `import rich_click as click` concept suggested by @willmcgugan in #8


ewels:
Hmm, did a speed back-of-the-car attempt just now @willmcgugan (I'm away travelling this weekend 😅) but no dice. Any ideas?

willmcgugan:
That looks like it should work. Were you getting any errors?

ewels:
No errors, but just default click behaviour - no customisation.

I need to import click but I guess the custom rich-click imports aren't overwriting those classes for some reason? I guess I could sort of do monkey-patching in this file to make it work 🤔

ewels:
eg. Running the [minimal example](https://github.com/ewels/rich-click/blob/import-as/examples/03_simple_importas.py), just get the default click output:

```console
$ python examples/03_simple_importas.py
Usage: 03_simple_importas.py [OPTIONS] COMMAND [ARGS]...

  My amazing tool does all the things.

  This is a minimal example based on documentation from the 'click' package.

  You can try using --help at the top level and also for specific group
  subcommands.

Options:
  -d, --debug / -n, --no-debug  Enable debug mode
  --help                        Show this message and exit.

Commands:
  download  Pretend to download some files from somewhere
  sync      Synchronise all your files between two places
```

I've tried a few things by trial and error, but I can't get the `rich-click` objects to be used with this method. I think I need a Python wizard with more skills than I have.. 👀

ewels:
Ok I can get it to work if do this in the `__init__.py`:

```python
from click import *
from .rich_click import rich_format_help

Group.format_help = rich_format_help
Command.format_help = rich_format_help
```

Then the `import rich_click as click` works nicely. Definitely an improvement for the end-user (less typing, less to get wrong), but it's still monkey patching - just in a different place. Which is kind of what we were trying to avoid with this...

I'll leave this open a little longer in case anyone has any ideas why the sequential import doesn't overwrite the class names in the way I expect, but if not I'll merge with this change.

grst:
I don't think the current approach can work, as `click` never get's to know that there's a new `Group` and `Command` class in `rich_click`. 

You could try something along the lines of:

`__init__.py`:

```python
from click import * 
from .rich_click import rich_format_help
from .rich_click import Group
from .rich_click import Command

def group(*args, cls=Group, **kwargs):
    from click import group as group_
    return group_(*args, cls=cls, **kwargs)
```

grst:
> We are again overwriting things in the click namespace, either way.

I wouldn't say so. My example only modifies the `rich_click` namespace.
So if a user did

```python
import click as cl
import rich_click as rcl
```

they could still use

```python
@cl.command()
def cli():
   pass

@rcl.command()
def cli2():
   pass
```

and one would be rich-click formatted and the other not. Whereas with monkey-patching, both would be rich-click formatted.

ewels:
> Whereas with monkey-patching, both would be rich-click formatted.

Is that true even when we're doing the monkey patching _within_ the `rich-click` package? As in https://github.com/ewels/rich-click/pull/10#issuecomment-1040870373 ?

ewels:
Damn, just tried it and you're right. Even when I do `Command.format_help = rich_format_help` within the `rich_click/__init__.py` it affects the class with the vanilla `import click as cl` import 😞

grst:
> However, if you try to get the help for a command within a group it uses the default click handling still.

There's a `command_class` attribute, it seems: 
https://github.com/pallets/click/blob/49164faca678dd476f1b11a2584fd0e1c6be70b2/src/click/core.py#L1757-L1762

Maybe it's possible to set it as keyword argument:
```python
return rich_group(*args, cls=cls, command_class=cls, **kwargs)
```

Otherwise it should be possible to override it in the `Group` class.

ewels:
ok awesome, just need to figure out how to handle the help now (`ClickException.show`). Need to run now..

ewels:
I'm not sure that we're going to be able to get at the error message printing code any way other than monkey patching the exception classes.. :/

Code is [nestled in the `main()` function](https://github.com/pallets/click/blob/c63c70dabd3f86ca68678b4f00951f78f52d0270/src/click/core.py#L1067-L1071):

```python
except ClickException as e:
    if not standalone_mode:
        raise
    e.show()
    sys.exit(e.exit_code)
```

I don't really fancy trying to replicate the massive `main()` function so I'm not sure that there's really any other way.. Unless anyone else has any smart ideas? 👀

grst:
Looks tough! I can't think of any better way either.

ewels:
Figured it out 💥 43842174adfa83a6945e61df474346142cb03359

Click hops about all over the place but managed to track back to the `main()` function call and override `standalone_mode` so that it allows the exceptions to raise. Then we can catch them with our custom code and handle the output nicely 🚀

PR #13, opened 2022-02-19
Conquer the land of clicketing snakes

pawamoy:
Twin sister of https://github.com/chriskiehl/Gooey/pull/788 :D
Original PR title: Allow to run any Click program from the CLI :p

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,8 +2,10 @@ from setuptools import setup
 
 setup(
     name="rich-click",
+    entry_points={"console_scripts": ["rich-click = rich_click.cli:main"]},
     install_requires=[
         "click",
         "rich",
+        "importlib-metadata; python_version < '3.8'",
     ],
 )
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,6 @@ from setuptools import setup
 
 setup(
     name="rich-click",
-    entry_points={"console_scripts": ["rich-click = rich_click.cli:main"]},
     install_requires=[
         "click",
         "rich",
--- a/setup.py
+++ b/setup.py
@@ -3,8 +3,8 @@ from setuptools import setup
 setup(
     name="rich-click",
     install_requires=[
-        "click",
-        "rich",
+        "click>=7",
+        "rich>=10",
         "importlib-metadata; python_version < '3.8'",
     ],
 )
--- a/src/rich_click/__main__.py
+++ b/src/rich_click/__main__.py
@@ -1,0 +1,15 @@
+"""
+Entry-point module, in case you use `python -m griffe`.
+
+Why does this file exist, and why `__main__`? For more info, read:
+
+- https://www.python.org/dev/peps/pep-0338/
+- https://docs.python.org/3/using/cmdline.html#cmdoption-m
+"""
+
+import sys
+
+from rich_click.cli import main
+
+if __name__ == "__main__":
+    sys.exit(main(sys.argv[1:]))
--- a/src/rich_click/__main__.py
+++ b/src/rich_click/__main__.py
@@ -13,4 +13,4 @@ from rich_click.cli import main
 
 if __name__ == "__main__":
     # main will run a Click command which will either exit or raise
-    main(sys.argv[1:])
+    main()
--- a/src/rich_click/__main__.py
+++ b/src/rich_click/__main__.py
@@ -12,4 +12,5 @@ import sys
 from rich_click.cli import main
 
 if __name__ == "__main__":
-    sys.exit(main(sys.argv[1:]))
+    # main will run a Click command which will either exit or raise
+    main(sys.argv[1:])
--- a/src/rich_click/__main__.py
+++ b/src/rich_click/__main__.py
@@ -1,13 +1,12 @@
 """
-Entry-point module, in case you use `python -m rich_click`.
+Entry-point module for the command line prefixer,
+called in case you use `python -m rich_click`.
 
 Why does this file exist, and why `__main__`? For more info, read:
 
 - https://www.python.org/dev/peps/pep-0338/
 - https://docs.python.org/3/using/cmdline.html#cmdoption-m
 """
-
-import sys
 
 from rich_click.cli import main
 
--- a/src/rich_click/__main__.py
+++ b/src/rich_click/__main__.py
@@ -1,5 +1,5 @@
 """
-Entry-point module, in case you use `python -m griffe`.
+Entry-point module, in case you use `python -m rich_click`.
 
 Why does this file exist, and why `__main__`? For more info, read:
 
--- a/src/rich_click/cli.py
+++ b/src/rich_click/cli.py
@@ -1,6 +1,7 @@
 """The command line interface."""
 
 import sys
+from textwrap import dedent
 from importlib import import_module
 
 try:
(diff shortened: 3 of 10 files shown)


pawamoy:
Just had to change a bit the imports in `__init__.py` to avoid infinite recursion :slightly_smiling_face:

pawamoy:
Any examples of projects using Click so I can test it on them? HTTPX seems to use its own formatting methods (same output with and without rich-click).

src/rich_click/cli.py
```diff
@@ -0,0 +1,46 @@
...
+
+import click
+from rich_click import group, command
+
+
+def main(args=None):
+    args = args or sys.argv[1:]
+    if not args:
+        # without args we assume we want to run rich-click on itself
+        # TODO: rewrite using argparse
```
pawamoy, line 18:
```suggestion
        # TODO: rewrite using click
```

src/rich_click/__main__.py
```diff
@@ -0,0 +1,15 @@
...
+- https://www.python.org/dev/peps/pep-0338/
+- https://docs.python.org/3/using/cmdline.html#cmdoption-m
+"""
+
+import sys
+
+from rich_click.cli import main
+
+if __name__ == "__main__":
+    sys.exit(main(sys.argv[1:]))
```
pawamoy, line 15: The `sys.exit` is probably not useful as I believe Click commands exit themselves?

setup.py
```diff
@@ -2,8 +2,10 @@
 
 setup(
     name="rich-click",
+    entry_points={"console_scripts": ["rich-click = rich_click.cli:main"]},
```
pawamoy, line 5: Not sure how to write this in setup.cfg, so here it is. Let me know if you'd still prefer it there.

src/rich_click/__main__.py
```diff
@@ -0,0 +1,15 @@
+"""
+Entry-point module, in case you use `python -m griffe`.
```
pawamoy, line 2:
```suggestion
Entry-point module, in case you use `python -m rich_click`.
```

pawamoy:
Well that's nice.

rich_uvicorn

src/rich_click/__main__.py
ewels: They do unless you set `standalone_mode` to `False`, then exceptions are raised (I've learnt a lot about click in the process of writing this project!)

ewels:
Brilliant stuff! 👏🏻 

I'm on holiday this week so won't get a chance to properly review + merge until after that, but love the idea / look of it..

Phil

src/rich_click/cli.py
```diff
@@ -0,0 +1,47 @@
...
+            del args[1]
+    sys.argv = [prog, *args[1:]]
+    # patch click before importing the program function
+    click.group = group
+    click.command = command
+    # import the program function
+    module = import_module(module_path)
+    function = getattr(module, function_name)
+    # simply run it: it should be patched as well
+    return function()
```
pawamoy, line 47: I don't pass any arguments to the function because it supports it in 100% of cases (that's how entrypoints work), while it will not work in 100% cases if I pass arguments (not mandatory, depends on each project).

pawamoy:
No problem, enjoy your holidays!

setup.py
ewels: Looks like it can be done: https://stackoverflow.com/a/48891252/713980

```
[options.entry_points]
console_scripts =
    rich-click = rich_click.cli:main
```

src/rich_click/__main__.py
pawamoy: Then I guess it's safer to keep the sys.exit here :slightly_smiling_face:

pawamoy: Oh wait, this is the main function of rich-click itself, so my first comment is irrelevant. I'll leave sys.exit since we use argparse.

ewels: Maybe - I sent this reply from my phone without seeing the code in context just before bed though, so it's not fully thought through 😅 But yeah, if it works without the `sys.exit` then I think it's probably fine to remove it 👍🏻

pawamoy: Wait again, I'm getting confused :joy: My first comment *is* relevant because the main function runs a Click function which in turn either exits or raises. So yeah, removing sys.exit :rofl:

ewels: confused

I'm a bit lost now 😆

pawamoy:
So there's just one more thing to decide on: whether to implement rich-click's own CLI with click or not.

ewels:
Yeah I did wonder about commenting about that! Seems a bit hypocritical not to somehow 😅

src/rich_click/cli.py
```diff
@@ -0,0 +1,51 @@
...
+        # without args we assume we want to run rich-click on itself
+        # TODO: rewrite using click
+        script_name = "rich-click"
+    else:
+        script_name = args[0]
+    scripts = {script.name: script for script in entry_points().get("console_scripts")}
+    if script_name in scripts:
+        # a valid script was passed
+        script = scripts[script_name]
+        module_path = script.module
```
ewels, line 27: Testing with [MultiQC](https://github.com/ewels/MultiQC) (before or after the point that I added rich-click support directly) I get this error:

```python
Traceback (most recent call last):
  File "/opt/miniconda3/envs/py3.8/bin/rich-click", line 33, in <module>
    sys.exit(load_entry_point('rich-click', 'console_scripts', 'rich-click')())
  File "/Users/phil/GitHub/ewels/rich-click/src/rich_click/cli.py", line 27, in main
    module_path = script.module
AttributeError: 'EntryPoint' object has no attribute 'module'
```

MultiQC has [a lot of entry points](https://github.com/ewels/MultiQC/blob/912f512d36459f147b1eac119c1bcea5f0594335/setup.py#L68-L210), which I guess is causing an issue..

ewels: I get the same error with other packages too, and in fact with `rich-click` itself:

```console
$ rich-click --help
usage: rich-click [SCRIPT | MODULE:FUNCTION] [-- SCRIPT_ARGS...]

$ rich-click
Traceback (most recent call last):
  File "/opt/miniconda3/envs/py3.8/bin/rich-click", line 33, in <module>
    sys.exit(load_entry_point('rich-click', 'console_scripts', 'rich-click')())
  File "/Users/phil/GitHub/ewels/rich-click/src/rich_click/cli.py", line 27, in main
    module_path = script.module
AttributeError: 'EntryPoint' object has no attribute 'module'
```

ewels (changes requested):
Ok, I had a stab at running locally but ran into some issues..

ewels:
Note to self: need to add the new entry points into the conda-forge `meta.yml` after this is released.

Example [here](https://github.com/conda-forge/svgdigitizer-feedstock/blob/bd7b00b29e03cafdcd7805a79f150f8b6dcf7d9d/recipe/meta.yaml#L13-L14), so something like this I guess:

```yaml
build:
  entry_points:
    - rich-click=rich_click.cli:main
```

src/rich_click/cli.py
pawamoy: Weird, `module` should always be available: https://docs.python.org/3/library/importlib.metadata.html#entry-points
What Python version are you using?

pawamoy: Ah, it's not in 3.8 indeed. I'll use `value` directly then.

ewels:
That fixed it! 🎉

ewels:
ok, so using click to handle the CLI was difficult because of the weird argument handling (we don't want click to interpret any option flags).

Instead, I tried to fake it - using as many of the rich-click functions and variables as possible so that the help text tracks with updates to the package.

This is what I now get:

image

I'll push this directly to your PR if that's ok  @pawamoy - I should PR to your feature branch really, but it's late.. 😁

ewels:
Now also spits out an informative error message if the command isn't found, or the full usage if just `rich-click` or `rich-click --help`:

image

ewels (approved):
Love it, working a treat in my hands now - thanks for this!

I don't think I'd have ever believed that this would work had you not put in this PR 😅

pawamoy:
> I'll push this directly to your PR if that's ok @pawamoy 

Totally fine!

And great work! I feared indeed that rewriting the CLI using click would be tricky, but you managed to do it well :smile:

PR #26, opened 2022-02-28
Add support for Typer

### LINKED ISSUE #14: Add support for Typer
This looks great! Thank you for sharing this.

I tried using the import in a Typer project and (unsurprisingly) it didn't work. Even if there's no official code support for using this with Typer, some docs would be super helpful since I would love to use it for my CLI.

browniebroke:
Fix #14 

Many thanks to @wasi-master for the original draft


paw-lu:
Thanks for this!

As this is, this change will break Typer's env var documentation.

Currently, if a typer Option has `envvar` set.

```python
typer.Option(
    False,
    "--bar",
    help="Lorep ipsum.",
    envvar="BAR",
)
```

Then it will automatically document it.

```console
% foo --help

Options:
--bar            Lorep ipsum.  [envvar: BAR]
```

Right now though, this change removes the automatic envvar documentation:

```console
% foo --help
╭─ Options ────────────────────────────────────────────────────────╮
│     --bar                                     Lorep ipsum.       │
│                                                                  │
```

paw-lu:
[click-contrib/click-help-colors](https://github.com/click-contrib/click-help-colors) is similar to this library, but preserves the envvar documentation when used with Typer ([it's what I'm currently using on a Typer project](https://github.com/paw-lu/nbpreview/blob/main/src/nbpreview/_color_typer.py#L27)). So it might be worth looking into the differences between what that and rich-click are doing.

ewels:
I think that the `envvar` thing is in click, so probably not related to the Typer work at all. Sounds like something I've missed (click does a tonne of stuff!) https://click.palletsprojects.com/en/8.0.x/options/#values-from-environment-variables

paw-lu:
> I think that the `envvar` thing is in click

Yeah, to be clear you're right—it does come from Click. Think the difference is that it's documented under `--help` by default in Typer, while for Click it is hidden by default (I think!).

> Made #36 to track this..

Thanks!

README.md
```diff
@@ -49,6 +49,22 @@ This means that you can continue to use the unmodified `click` package in parall
 
 > See [`examples/02_declarative.py`](examples/02_declarative.py) for an example.
 
+## Typer support
+
+[`Typer`](https://github.com/tiangolo/typer) is also supported. You may want to change the installation to use the the `typer` extra:
```
ewels, line 54:
```suggestion
[`Typer`](https://github.com/tiangolo/typer) is also supported.
You need to use rich-click with the `typer` [extra](https://packaging.python.org/en/latest/tutorials/installing-packages/#installing-setuptools-extras) in your package requirements: `rich-click[typer]`

For example, to install locally:
```

README.md
```diff
@@ -49,6 +49,22 @@ This means that you can continue to use the unmodified `click` package in parall
...
+
+```bash
+python -m pip install rich-click[typer]
+```
+
+And replace your usual `typer` import by:
```
ewels, line 60:
```suggestion
Then just replace your usual `typer` import by:
```

ewels (commented):
Hi all,

I've just had another read over the code and I think it looks great - love how simple it is 👌🏻

I was partly waiting to see if there would be imminent movement on the core Typer library to include support. Sounds like [it might happen](https://twitter.com/tiangolo/status/1498512578649067520?s=20&t=3SEmCu6mAaB1_w6_uy0F5g), but I figure that there's no problem in having both and we can always strip this out again at a later date if it becomes redundant.

Just a couple of minor readme suggestions for additional clarity about installation requirements, then the crash that I'm getting when I run the example script.

Phil

browniebroke:
> then the crash that I'm getting when I run the example script.

If I remember well, I was getting the same crash with some of the Click examples. I assumed it was my environment that was messed up.

examples/08_typer.py
```diff
@@ -0,0 +1,56 @@
...
+    This is a minimal example based on documentation
+    from the 'click' package.
+
+    You can try using --help at the top level and also for
+    specific group subcommands.
+    """
+    print(f"Debug mode is {'on' if debug else 'off'}")
+
+
+@cli.command()
```
ewels, line 20:
```suggestion
@app.command()
```

examples/08_typer.py
```diff
@@ -0,0 +1,56 @@
...
+def sync(
+    type: str = typer.Option("files", help="Type of file to sync"),
+    all: bool = typer.Option(False, help="Sync all the things?"),
+):
+    """Synchronise all your files between two places"""
+    print("Syncing")
+
+
+@cli.command()
```
ewels, line 29:
```suggestion
@app.command()
```

examples/08_typer.py
```diff
@@ -0,0 +1,56 @@
...
+
+    \f
+    Also if you want to write function help text that won't
+    be rendered to the terminal.
+    """
+    print("Downloading")
+
+
+if __name__ == "__main__":
+    cli()
```
ewels, line 56:
```suggestion
    app()
```

ewels (commented):
I replicated the error on another computer, so not sure that it's just my environment (though the setups are similar so there could be something odd).

After a bit of trial and error testing out examples from the Typer docs (which worked), I came up with the three suggested changes here which make the example script work for me.

Phil

ewels:
Whilst going through the Typer docs (which I'm unfamiliar with, I've not used Typer for a project before), I noticed that the basic examples don't get any rich-click functionality. For example, [the simplest example](https://typer.tiangolo.com/tutorial/first-steps/#the-simplest-example):

```python
import rich_click.typer as typer

def main():
    typer.echo("Hello World")

if __name__ == "__main__":
    typer.run(main)
```

Gives no rich output:

```console
$ python test_typer.py  --help
Usage: test_typer.py [OPTIONS]

Options:
  --install-completion [bash|zsh|fish|powershell|pwsh]
                                  Install completion for the specified shell.
  --show-completion [bash|zsh|fish|powershell|pwsh]
                                  Show completion for the specified shell, to
                                  copy it or customize the installation.
  --help                          Show this message and exit.
```

The rich-click functionality only seems to click in when using the decorator functionality:

```python
app = typer.Typer()
@app.command()
def cli():  #...
```

I don't know enough to speculate why this is, or how we could support the other usage method. However, we should probably mention this in the readme at least to prevent confusion.

Phil

ewels:
Sounds good, thanks! Happy to merge this PR now and add the additional support for the basic examples in a later PR if you prefer, just let me know.. (also re: my suggested example code edits above)

Phil

browniebroke:
I've updated the example, it shoudl now work. On top of your suggestions, I've also:

- renamed the file to not have 2 `examples/08_...`
- changed the decorator of `cli` from `@app.command()` to `@app.callback()`: without that, the `cli` was detected as a subcommand rather than the help of the CLI itself. It also enables the top level `--debug` option to work, similar to what is done in the `01_simple.py` example

src/rich_click/typer.py
```diff
@@ -0,0 +1,33 @@
...
+        **kwargs,
+    ) -> Callable[[CommandFunctionType], CommandFunctionType]:
+        return super().command(*args, cls=cls, **kwargs)
+
+
+def run(function: Callable[..., Any]) -> Any:
+    """Redefine typer.run() to use our custom Typer class."""
+    app = Typer()
+    app.command()(function)
+    app()
```
browniebroke, line 33: That's copied from typer itself:

https://github.com/tiangolo/typer/blob/6d8e954664d433fb8beb63cd1329f21554ea7b36/typer/main.py#L861-L864

I don't see a better way to inject our `Typer` class, open to suggestions.

ewels:
> I'm thinking to move the typer example in a dedicated directory, which would enable adding more than 1, thoughts?

Yes, I was thinking the same - have directories for click and typer. Maybe can drop the numeric prefixes too, as they're getting annoying. It made sense when there were about 3 as it helped walk through them logically, but as their number grows it's a hassle.

They're sort of doubling as crap tests at the moment, so hopefully once we get proper unit testing in place they might not proliferate quite so quickly! 😅

ewels:
Sorry for all of the commit spam with the new CI precommit tests. They've been passing locally for me which made it a bit tricky to work with. Anyway, they were getting annoying for the examples so I've just set flake8 and mypy to ignore that folder. Now it's passing so I think we're good..

browniebroke:
No worries, your updates look good to me :+1:thanks for tyding it up!

ewels (approved):
Thanks again for this! Let's go for it and brace for the issues that start coming in once it's supported 😅

browniebroke:
Interesting timing... FYI A new version of click was released today and it broke Typer: https://github.com/tiangolo/typer/pull/375 (in case you get one of these issus)

ewels:
hah - good to know, thanks!

tiangolo:
Hey, hey! I just released Typer `0.4.1` handling the incompatibility with Click 8.1.0. :rocket: :nerd_face:

ewels:
Hi all - I just added support for click `envvar` (which also gives support for Typer `envvar`). Thanks for bringing this to my attention @paw-lu! See https://github.com/ewels/rich-click/issues/36 for details (but it should "just work").

PR #37, opened 2022-03-07
Move RichGroup and RichCommand out of rich_click.py

Jorricks:
First of a couple refactorings that are coming up.
This refactoring:
- Removes local imports. It's generally more preferred to use absolute imports.
- I moved the classes out of rich_click.py. This made the rich_click.py unnecessarily large..
- Moved the creation of the console into a function, as it's used all over the place.

I think to tune this file down more, we might consider moving all rich specific tools to a file called `rich_utils` and moving the click items into `click_utils`.

### DIFF
--- a/src/rich_click/__init__.py
+++ b/src/rich_click/__init__.py
@@ -11,8 +11,8 @@ __version__ = "1.2.2.dev0"
 from click import *
 from click import group as click_group
 from click import command as click_command
-from .rich_click import RichGroup
-from .rich_click import RichCommand
+from rich_click.rich_group import RichGroup
+from rich_click.rich_command import RichCommand
 
 
 def group(*args, cls=RichGroup, **kwargs):
--- a/src/rich_click/rich_click.py
+++ b/src/rich_click/rich_click.py
@@ -10,7 +10,6 @@ from rich.table import Table
 from rich.text import Text
 from rich.theme import Theme
 import re
-import sys
 
 # Support rich <= 10.6.0
 try:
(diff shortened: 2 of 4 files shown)


Jorricks:
Other things I was thinking would be a good idea:
1. Add typing info to each argument of a function
2. Introduce pre-commit and thereby: [black](https://github.com/psf/black), [flake8](https://github.com/PyCQA/flake8) and [pytype](https://github.com/google/pytype).
What do you think @ewels?

ewels (commented):
Yes, much better - thanks 👍🏻  The console thing had been in the back of my mind for a bit as well.

I wonder if we can just create a single top-level global console object and then use that everywhere? (Similar to the `highlighter`). I'm not sure that there's any need to create so many different versions of the same thing. Might be cleaner.

ewels:
> Other things I was thinking would be a good idea:
> 
> 1. Add typing info to each argument of a function
> 2. Introduce pre-commit and thereby: [black](https://github.com/psf/black), [flake8](https://github.com/PyCQA/flake8) and [pytype](https://github.com/google/pytype).
>    What do you think @ewels?

Yes to all of this 👍🏻  I'm a scientist first and coder second and have never gotten around to properly teaching myself to use typing. But I've been feeling increasingly guilty about it! If you fancy adding typing to this then that would be great actually, as it'll force me to use it for any new features I add 😅 

Pre-commit stuff also good. Again I've never bothered setting this up myself (instead relying on VSCode plugins) but I should probably set myself up to use it.

Jorricks:
> Yes, much better - thanks 👍🏻 The console thing had been in the back of my mind for a bit as well.
> 
> I wonder if we can just create a single top-level global console object and then use that everywhere? (Similar to the `highlighter`). I'm not sure that there's any need to create so many different versions of the same thing. Might be cleaner.

I don't like global variables to be far.
I think for the way you are using them for settings, it's quite nice, but in general for things like getting a console, I rather use functions. I usually achieve this with the `functools.lru_cache` decorator. 
I think it would fit well in this use-case as well.

Jorricks:
Fixed black formatting 👍 .
Cool I'll proceed with the other PRs once this one is concluded :v:

ewels (approved):
Changelog entries would be good, but as you're going to have a follow-on PR can do that later.

PR #39, opened 2022-03-08
Add typing info

Jorricks:

PR #40, opened 2022-03-08
DRAFT Add pre-commit

Jorricks:
Added the following tools to pre-commit:
- iSort (import formatter)
- Black (general formatter)
- Flake8 (style checker)
- Mypy (type checker)

Furthermore, modified the README.md to contain new instructions.
Two more things are left to do:
- [ ] Run pre-commit on the entire codebase.
- [ ] Add item to changelog.

PR #41, opened 2022-03-08
Add PEP 561 Compatiblity

harens:
Hi there 👋 

I just saw https://github.com/ewels/rich-click/pull/39 and thought it might be useful to add a `py.typed` file. This allows downstream projects to use the type hints for things like PyCharm, mypy, pytype etc.

Without this file, mypy currently complains in projects that use rich-click:

```
error: Skipping analyzing "rich_click": module is installed, but missing library stubs or py.typed marker
```

The `py.typed` file makes rich-click [PEP-561 compatible](https://www.python.org/dev/peps/pep-0561), and allows the inline types to be used. See [here](https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/) for further explanation.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,4 +7,5 @@ setup(
         "rich>=10",
         "importlib-metadata; python_version < '3.8'",
     ],
+    package_data={"rich-click": ["py.typed"]},
 )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Version 1.2.2.dev0
 
+- Mark PEP 561 Compatibility [[PR41](https://github.com/ewels/rich-click/pull/41)]
 - Add typing information [[PR39](https://github.com/ewels/rich-click/pull/39)]
 - Refactor RichCommand and RichGroup out of rich_click [[PR38](https://github.com/ewels/rich-click/pull/39)]
 - Change metavar overflow to `fold`, so that large numbers of choices flow onto new lines instead of being truncated with an ellipsis [[#33](https://github.com/ewels/rich-click/issues/33)]
--- a/src/rich_click/py.typed
+++ b/src/rich_click/py.typed
@@ -1,0 +1,1 @@
+# Marker file for PEP 561. rich-click uses inline types.


ewels:
Nice! Yeah, @Jorricks has been helping with this as I am new to Python typing.. 👴🏻

Should we be running mypy with GitHub Actions to check for errors in the typing?

harens:
Thank you @ewels for merging, and thanks @Jorricks for your effort on the typing front. 👍 I know from experience just how much work that can be, so I really appreciate what you're doing.

> Should we be running mypy with GitHub Actions to check for errors in the typing?

> Regarding running MyPy, that's included in the pre-commit.

Do you trust that the developer will always install the pre commit? ;)

As for mypy in general, I'd say it's quite a double-edged sword. It's saved me from mistakes in the past, but it takes a lot of effort to squash out errors from `mypy`, let alone `mypy --strict`. You might prefer the [beartype](https://github.com/beartype/beartype) approach of checking types at runtime:

[From](https://github.com/beartype/beartype#when-should-i-use-beartype) @leycec:

> You want to write code rather than fight a static type checker, because [static type inference](https://en.wikipedia.org/wiki/Type_inference) of a [dynamically-typed](https://en.wikipedia.org/wiki/Type_system) language is guaranteed to fail and frequently does. If you've ever cursed the sky after suffixing working code incorrectly typed by [mypy](http://mypy-lang.org/) with non-portable vendor-specific pragmas like # type: ignore[{unreadable_error}], beartype was written for you.

ewels:
> Do you trust that the developer will always install the pre commit? ;)

The GitHub Actions CI workflow runs the pre-commit checks, so they don't need to - the PR checks will fail if it doesn't validate. See #40

Jorricks:
Well in my opinion, MyPy is indeed a tiny bit annoying but it prevents so many bugs. 
Furthermore, it makes sure you are set up for success around None values. And so much more.. I always felt it is worth the extra effort.

setup.py
```diff
@@ -7,4 +7,5 @@
         "rich>=10",
         "importlib-metadata; python_version < '3.8'",
     ],
+    package_data={"rich-click": ["py.typed"]},
```
ewels, line 10: Bit late now, but maybe we could put this into `setup.cfg` instead? Most stuff is in there, it's only the requirements that are in `setup.py` so that they can be picked up by GitHub:

https://github.com/ewels/rich-click/blob/6d444bf9dd95c5b93e68da9fac695d4914116dbc/setup.cfg#L39

PR #42, opened 2022-03-08
Add MacPorts Distribution

harens:
I've found this project really useful, so I thought that I could help it most by adding it to [MacPorts](https://www.macports.org/). You can find it's project page [here](https://ports.macports.org/port/py-rich-click/), and it can be installed on macOS by running the following:

```
sudo port install py-rich-click
```

Finally, thank you so much @ewels for keeping this amazing project going. It has been such a useful library and I really appreciate the work you've put into it. 👍

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Version 1.2.2.dev0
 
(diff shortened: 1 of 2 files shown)


ewels (approved):
Nice, thank you very much! 👏🏻

PR #43, opened 2022-03-14
Override the class in addition to the decorator.

dwreeves:
This PR fixes an issue wherein many popular libraries' CLIs (such as Celery and Flask) do not work as intended with `rich-click [command]`.

The reason why the current approach doesn't work is because these libraries opt to subclass `click.Group` instead of using decorators for the entrypoints `flask` and `celery`. In these cases, patching `click.Command` instead of `click.command` does the trick.

However, it is still necessary to patch the decorators in other cases, e.g. `rich-command flask --help` needs `click.Group` (class) to be patched, but `rich-command flask run --help` needs `click.command` (decorator) to be patched.

### DIFF
--- a/src/rich_click/__init__.py
+++ b/src/rich_click/__init__.py
@@ -5,7 +5,7 @@ The intention is to provide attractive help output from click, formatted with ri
 customisation required.
 """
 
-__version__ = "1.2.2.dev0"
+__version__ = "1.2.2.dev1"
 
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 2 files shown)


ewels:
Super nice, thanks! 👏🏻

dwreeves:
oh woops. Sorry about not running the pre-commit; I was experiencing an error message. 😅 In retrospect I should have googled the error message. Also, I didn't update the readme for the version bump.

ewels:
No worries! And I haven't usually bothered bumping the dev version as I go along yet so not too fussed about the changelog either 🤠 Feel free to put in another quick PR if it bothers you 😅

PR #46, opened 2022-03-18
fix #45

### LINKED ISSUE #45: Negative `--no-*` option variant not showing up
It seems that the `--no-*` variants for options are not showing up for me... I took the same example from #31 and it produces the following result when I run it:

image

These are the versions of `rich` and `rich-click` installed in the virtualenv:
- `rich`: 10.16.2
- `rich-click`: 1.2.1

taranlu-houzz:
Fixes #45. Not sure if there is more to be considered here, since I am not really familiar with the inner workings of `click`.


ewels:
Great stuff, thanks! I have a feeling I put the idx stuff in because there can be multiple pairs of these flags (you can define lots of alternatives for the same option) and we need to show the right one. But if this fixes the behaviour then I guess you've isolated where the bug is 👌🏻 I'll take a look ASAP.

ewels (approved):
Thanks again for this! I refactored slightly to use a `try`/`except` with the assumption that this is the kind of safety mechanism that I was aiming for when I originally wrote the code.

In my testing I couldn't get it to trigger so I'm not sure that it is needed, but I figure it doesn't hurt either.

PR #51, opened 2022-03-27
Support pre-processing Emoji codes in markdown. USE_MARKDOWN_EMOJI

### LINKED ISSUE #50: Use of emojis in Markdown
Rich does not support emoji in markdown out of the box.
But there is a simple workaround:

```python 
from rich.emoji import Emoji
markdown = "# Hello World :thumbs_up:"
markdown = Emoji.replace(markdown)
```

Taken from https://github.com/Textualize/rich/issues/231

schettino72:
Fixes #50.


ewels:
Great stuff, thanks! Is there a time when people wouldn't want this enabled? I'm wondering if we even need a variable for it (or at least if it should be on by default).

schettino72:
> Great stuff, thanks! Is there a time when people wouldn't want this enabled?

My reasoning was that this is non-default behaviour from `rich`...

> I'm wondering if we even need a variable for it (or at least if it should be on by default).

Yes, I was wondering the same... I took the more conservative approach but not against it. your call :smile:

ewels:
I can only think of one or two situations where this could give unexpected results, mainly very short emoji codes like `:x:` ❌ and one or two others..

I think it's fine to go for what the majority will expect and have this as the default. Then if it's causing problems then people can turn it off 👍🏻

ewels:
Great stuff, thanks! 🙌🏻

PR #58, opened 2022-04-18
Add autocomplete for typer

Kludex:
The `Typer` in the image below is from this PR. Installed via `rich-click @ git+https://github.com/Kludex/rich-click@add-autocomplete-typer`.

image

The idea of this PR is to improve the autocompletion on the `rich_click.typer` proxy.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -2,3 +2,5 @@
 # Default ignores that we can extend
 ignore=D100,D102,D205,E203,E231,E731,W504,I001,W503
 max-line-length=120
+per-file-ignores=
+    typer.py:F401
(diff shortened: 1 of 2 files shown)


ewels:
Whilst I see the point of this, I can't say that I love the change. It makes the code much messier and I can imagine it being fairly easy to get errors triggered by variations of installed package versions (when we don't actually care about all of these arguments).

Is there really no other way to achieve this? 😕 I'm a bit on the fence as to whether the benefits outweigh the disadvantages here.

ewels:
I think I'll close for now if that's ok. I'm worried that the burden of maintaining this will be too great, and at the end of the day it's for a mostly-cosmetic advantage. We can revisit if lots of other people start requesting it (note that hopefully Typer will have built-in support for this kind of CLI output already so this package may become redundant for Typer).

PR #62, opened 2022-04-27
add support of hidden command option

lgprobert:


Benjamin-Lee:
Great! Thanks! I've needed this

ewels (approved):
LGTM, thanks!

ewels:
Ok, I was briefly confused as in the issue you said that this was a click 8.1 feature. But actually it goes back 6 years to version 7.0 😅  So no need to cater for it not existing (which is what I was trying to test).

Issue #55, opened 2022-04-13
`rich_cli` does not properly clean indentations for `click >= 8.1.0`


apcamargo:
`rich_click` is not cleaning indentations if `click` version is ≥ 8.10. The cause is probably [this change](https://github.com/pallets/click/issues/1985) in the backend.

This code:
```python
def cli():
    """
    Foo: bar

    Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod
    tempor incididunt ut labore et dolore magna aliqua. Pellentesque habitant
    morbi tristique senectus. Leo a diam sollicitudin tempor.
    """
```

Is generating this:
```
 Foo: bar
 Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod     tempor incididunt ut
 labore et dolore magna aliqua. Pellentesque habitant     morbi tristique senectus. Leo a diam
 sollicitudin tempor.
```

Instead of this:
```
 Foo: bar
 Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore
 et dolore magna aliqua. Pellentesque habitant morbi tristique senectus. Leo a diam sollicitudin
 tempor.
```

ewels:
Ok, so we need to run dedent manually now I guess. Thanks for reporting! PRs welcome, I'm away for 2-3 weeks now with holiday and starting a new job but will get to it ASAP.

apcamargo:
I'll try to take a look at the code in the next few days. But it seems to me that the easy way out would be to use `inspect.cleandoc` within `_make_rich_rext`. Do you think this would beak something?

ewels:
No, that sounds perfect 👍🏻  I wasn't familiar with `inspect.cleandoc`, I didn't realise that this was part of the core lib. I've used `textwrap.dedent` in the past which is why that jumped to mind. But better to use the same approach as Click 👍🏻

abdusco:
@apcamargo 's fix works wonderfully 🚀

apcamargo:
@ewels I'm still having this issue in the latest version :(
@abdusco, how did you test the PR?

apcamargo:
Ok. Apparently the fix worked for the header, but not for the command description.

```
 Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut
 labore et dolore magna aliqua. Pellentesque habitant morbi tristique senectus. Leo a diam
[...]

PR #67, opened 2022-05-15
Make `_make_command_help` remove text indentations with `inspect.cleandoc`

apcamargo:
My last PR didn't fix https://github.com/ewels/rich-click/issues/55 properly. The command help texts were not being sanitized.

I tested this PR and this time it really fixes the issue.


ewels (approved):
Good catch, sorry I missed this. Thanks for the fix!

PR #78, opened 2022-06-21
Update build

zmoon:
* Use `python -m build` instead of invoking `setup.py`, in accordance with "SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools."
* Update Action versions
* Minimize what is included in `setup.py` (only the deps, to get in the Dependency Graph)


ewels (approved):
Nice, thanks!

zmoon:
Thanks for this package, I've been enjoying using it!

PR #79, opened 2022-06-21
Fix #63 for RichGroup

### LINKED ISSUE #63: Return value from Typer function is printed only when installed
I apologize in advance for how obscure this bug is.

I'll start by taking the simple Typer demo and modifying it to have `download` return a string:

```python

import rich_click.typer as typer

app = typer.Typer()

@app.command()
def sync(
    type: str = typer.Option("files", help="Type of file to sync"),
    all: bool = typer.Option(False, help="Sync all the things?"),
):
    print("Syncing")


@app.command()
def download(all: bool = typer.Option(False, help="Get everything")):
    return "Shouldn't appear" # <-- RETURNS A STRING NOW


if __name__ == "__main__":
    app()
```

When run directly, everything is normal:

```
$ python vdsearch/test.py download
$
```

Now let's `pip install -e .` with `testy` mapped to `test:app`

```
$ testy download
Shouldn't appear
$
```

I've gotten around this bug by defining wrapper functions for my functions but it would be nice to get to the bottom of what's going on.

mcflugen:
@ewels Oops! In #76 I applied the fix to `RichCommand` but not also to `RichGroup`. 🤦‍♂️ Anyway, this pull request fixes the same issue described in #63 but for `RichGroup`.

### DIFF
--- a/src/rich_click/rich_group.py
+++ b/src/rich_click/rich_group.py
@@ -18,7 +18,9 @@ class RichGroup(click.Group):
 
     def main(self, *args, standalone_mode: bool = True, **kwargs):
         try:
-            return super().main(*args, standalone_mode=False, **kwargs)
+            rv = super().main(*args, standalone_mode=False, **kwargs)
+            if not standalone_mode:
+                return rv
         except click.ClickException as e:
             if not standalone_mode:
                 raise


ewels (approved):
Sorry, I should have caught this too. Thanks for spotting / fixing!

mcflugen:
@ewels Any chance we could get a *v1.5.1* bug-fix release that includes this PR? Thanks!

ewels:
Sure, done: https://pypi.org/project/rich-click/1.5.1/

Issue #89, opened 2022-10-03
Make `rich-click` extensible and isolate configuration state


BrutalSimplicity:
Right now all of the functions in the `rich_click` module are difficult to override and extend behavior unless you extend one of the `RichCommand` or `RichGroup` classes. And even then, you can't really get access the the internals since the `formatter` is not being used and the results of `rich_format_help` are not returned to the caller.

It would be nice if the Rich `Console` and `rich_click` configuration state could be configured/extended through class inheritance.

This leads to two implementation changes.
- One, move this behavior into a class with the configuration state encapsulated in the class.
- Two, create extensible methods that allow one to change the `Console` configuration and/or results of certain formatting functions.

For the first item, I think it might be possible to provide a class implementation of the formatter that could contain the state, and then expose a new method on the `Context` object that would allow you to configure the formatter before making the call to `rich_format_help`. For the most part, the module state is not a big problem since help generation is fired only once before application termination. However, one issue I've found is that letting the configuration sit in module state makes testing help messages difficult, since state changes leak across tests. I think it's also generally good practice to isolate state, and will make this library more reusable across the `Click` ecosystem.

I think the changes for one would solve most of the issues with two. If you implement the `formatter` interface, others could hook into that functionality, and it could be configured along with whatever method or attribute you expose on the `Context`.

ewels:
Sounds sensible 👍🏻  Do you think it would be possible to do this in a backwards-compatible manner for anyone who is already extending those classes?

BrutalSimplicity:
I took a stab at this yesterday evening. I'll submit my PR for initial review later today.

BrutalSimplicity:
Opened a PR for this that coincidentally seems to tackle quite a few other issues

#25 #90 #88 #29 #18

PR #92, opened 2022-10-23
Feat #89 isolation and extensibility design

BrutalSimplicity:
This is a large PR, and for that I apologize. I'm happy to make changes and provide additional explanation based on feedback.

The main focus of this PR is to add support for isolating Rich Click's formatting and help configuration. The design supports isolation of these components by:

1. Extending Click's `HelpFormatter` class
2. Creating a `HelpConfiguration` class that doubles the current module-level settings
3. Adding a decorator that allows the `HelpConfiguration` to be passed into Click via the supported `context_settings` argument provided by the Command and Group classes.

Since this package does have an established user base I attempted to maintain the module-level settings behavior to prevent any breaking changes. I also added some tests to help verify the test suite using `pytest` and `pytest-cov` for test coverage. The tests are the bulk of this change as I used a snapshot-based approach to capture the input and output of the examples, and then validate their compatibility between module-level and context-level configuration.

I got it a bit carried away and added a few other nuggets since they seemed closely related.

1. The Rich Console object can also be configured per command and is distinct from the Console instance used internally by the formatter. The `RichHelpFormatter` creates a console based on the `RichHelpConfiguration` as the tight coupling between the Formatter and Click's internals make it difficult to allow the Console to be configured externally (i.e. one example is that Click expects help formatting to be buffered).
2. `RichContext` class was created to allow creation of the custom formatter.
3. The Rich Command, Group, and Context now expose the `Console` and `RichHelpConfiguration` properties. I thought this might make sense, but happy to remove if it might cause problems elsewhere.
4. Added contributor vscode settings


BrutalSimplicity:
The `rich-codex` workflow updated the images, but seems like the screen grabs are no longer generating colorful output for some of the commands. Not quite sure what to make of it. Maybe some leaky module config state carrying over across snapshots? When I run `rich-codex` or the `rich-click` examples I always seem to get some unsupported characters in my terminal.

image

.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+  "python.defaultInterpreterPath": "venv/bin/python",
```
ewels, line 2: Do we need this file here? I usually try to avoid having `.vscode` config files in repos, as many of these things can come down to personal taste. For example, I don't use `venv` for Python myself - I use `conda` (the most common toolchain in my field).

Happy to be won over though.

ewels:
This is amazing, thanks so much! A tonne of stuff to go through so it may take me a few days to find time to properly look at it, but on the face of it I think it looks great 🚀

ewels:
Note - `rich-codex` running `on: [push]` is nice in principle but in practice I've been finding it super spammy on other projects. So feel free to switch it to just `workflow_dispatch:` if you like. Can then try to remember to run it before release.

.vscode/settings.json
BrutalSimplicity: That's a good point. What do you think about converting that to something like, `settings.template.jsonc`, with some useful commentary on the settings for those that want to opt-in?

`jsonc` (json with comments) is a supported format in vscode that allows adding commentary to json files.

BrutalSimplicity:
Yea, I hear you on the `rich-codex` workflow.

I made a fix to handle the issue I found with improper usage formatting. It looks like internally click creates a new formatter to print out the usage string, so the Rich formatting was being applied twice. Now, instead of asking click for the usage I changed that portion to call `write_usage` on the formatter and that seemed fix the behavior.

Other than that, I _believe_ (please verify) that any other changes were only updates to configuration attributes and threading the formatter through to other functions.

Perhaps, it might be worth looking at moving that functionality into the formatter itself, but the PR is large enough as is, and I didn't want to accidentally break the internals.

src/rich_click/rich_command.py
```diff
@@ -13,21 +16,61 @@ class RichCommand(click.Command):
...
+        """Rich Help Formatter.
+
+        This is separate instance from the formatter used to display help,
+        but is created from the same `RichHelpConfiguration`. Currently only used
+        for error reporting.
+        """
+        return self._formatter
 
     def main(self, *args, standalone_mode: bool = True, **kwargs):
+        formatter = self._formatter = RichHelpFormatter(config=self.help_config)
```
BrutalSimplicity, line 58: Not sure the formatter is worth exposing here. It's only needed in the exception handling at this point as the formatter used for help generation is created by click internally. I only exposed it as a possible convenience for users. Maybe the Rich console is enough.

src/rich_click/rich_click.py
```diff
@@ -99,54 +106,52 @@
...
-    """Highlights our special options."""
 
-    highlights = [
-        r"(^|\W)(?P<switch>\-\w+)(?![a-zA-Z0-9])",
-        r"(^|\W)(?P<option>\-\-[\w\-]+)(?![a-zA-Z0-9])",
-        r"(^|\W)(?P<argument>[A-Z0-9\_]+)(?![_a-zA-Z0-9])",
-        r"(?P<metavar>\<[^\>]+\>)",
-        r"(?P<usage>Usage: )",
-    ]
+def _get_rich_formatter(formatter: Optional[click.HelpFormatter] = None) -> RichHelpFormatter:
```
BrutalSimplicity, line 116: This is possibly not needed since the formatter entry points are in the public functions. But, I was a bit unsure of other ways users may be using the "private-by-convention" functions in this module.

My thought was maybe someone is calling these internal functions to supplement behavior, and are calling them without the newly added `formatter` argument. In that case, this should fall back to the original behavior since it will then resolve the formatter from the active click context or the module-level configuration.

.vscode/settings.json
ewels: Sure, feels slightly overkill to be but wouldn't do any harm, so happy to have that if you'd like it 😀 

> `jsonc` (json with comments) is a supported format in vscode that allows adding commentary to json files.

Awesome - never heard of this file format before! I always just use YAML instead if I can 😅

ewels:
Hi @BrutalSimplicity,

Sorry for being slow on the review here. If I'm honest, much of what you've written is a bit beyond me 😅  I'm keen to merge, but it makes me nervous about maintaining the code if I don't really understand it. What do you think about being added as a collaborator on the repo to help out more long term? Then I would feel much happier to merge even though I'm out of my depth.

General points:

- Screenshots in the PR are still lacking colour, even if fixed in the workflow. Can probably just do a bit of rebasing to remove those commits? I've switched to `workflow_dispatch` outside of this PR already.
- Would be great to add some more docs / examples if we're adding new extensibility here.

Cheers,

Phil

PR #98, opened 2022-11-14
Fix click command call

davidbrochart:
Follow-up of #95.

PR #103, opened 2023-03-28
Feat #89 isolation and extensibility design

ewels:
Continuation of https://github.com/ewels/rich-click/pull/92 in a new PR, so that I can force-push to remove some commits.


ewels:
@davidbrochart - I'm having issues updating this PR due to conflicts in `__init__.py` with the changes that you made in https://github.com/ewels/rich-click/pull/98

Are you able to help me figured out what your command syntax should look like with this new `wrapper()` call please?

davidbrochart:
I tried in #104, but not sure of myself.

src/rich_click/__init__.py
```diff
@@ -32,22 +37,93 @@
...
+        else:
+            group = click_group(name, cls=cls, **attrs)
+        cmd = cast(RichGroup, group(fn))
+        return cmd
 
-def command(*args, cls=RichCommand, **kwargs):
+    return wrapper
+
+
+def command(*args, cls=RichCommand, **kwargs) -> Callable[..., RichCommand]:
```
ewels, line 72: @davidbrochart do we need to update `command` as well as `group` in the same way as done in #104?

ewels: Ok, I basically copied the change you made and applied it to this function as well (modifying where relevant to be command instead of group). Hope that's ok 🤞🏻 

I was never able to replicate the error that you were fixing in the first place, so I'll leave it to you to shout if it's broken in any way again 😁

ewels:
ok, I'm merging this before any more merge conflicts crop up.

I'm still not totally clear on what all of the code changes do, but I am pretty sure that it's a major improvement over the current state. If people find any problems then hopefully we can sort them relatively fast.

Massive thanks to @BrutalSimplicity for this work! 🙇🏻

src/rich_click/__init__.py
davidbrochart: Indeed, thanks! There was just a small typo (see #105).

ewels: Nice catch - thanks! 🙌🏻

PR #104, opened 2023-03-28
Fix group

davidbrochart:


ewels:
Failing test seems to be due to a rich update (was working locally until I tried upgrading)

PR #105, opened 2023-04-12
Replace click_group with click_command

davidbrochart:


ewels (approved):
ooh, that was close - good catch! Thanks!

Issue #18, opened 2022-02-23
options inherited from context settings aren't applied


seanbreckenridge:
Thanks for the library, love that its a drop in replacement

Just creating this to track this here/let you know. 

Defaults set in [context settings](https://click.palletsprojects.com/en/8.0.x/api/#click.Command) aren't detected by command/groups

Minimal example:

```python
import os

USE_RICH = "USE_RICH" in os.environ

if USE_RICH:
    import rich_click as click  # type: ignore[import]
else:
    import click  # type: ignore[no-redef]

DEFAULT_CONTEXT_SETTINGS = {"show_default": True}


@click.command(context_settings=DEFAULT_CONTEXT_SETTINGS)
@click.option("--use-context-settings", default="should work", help="help text shows")
@click.option("--overwrite-to-succeed", default="does work", show_default=True, help="shows here too")
def main(use_context_settings, overwrite_to_succeed) -> None:
    click.echo(use_context_settings)
    click.echo(overwrite_to_succeed)

if __name__ == "__main__":
    main()
```

![image](https://user-images.githubusercontent.com/7804791/155363615-be3b11fb-0268-4ecf-b00a-6ac7c6a9c26b.png)

Would be willing to create a PR for this at some point in the future, I would imagine one just has to look these up in the parent context instead of just checking the parameters passed, but may be a bit more complicated than that...

ewels:
Ah well spotted! Many thanks for the minimal example too.. A PR would be great if you get a chance (I'm away on holiday currently) otherwise I'll try to look into it when I get a chance 👍🏻

seanbreckenridge:
Tried [creating a basic helper](https://github.com/seanbreckenridge/rich-click/blob/6aa3d0b2ccbb5535827c60ac4a9f6de02aa027e6/src/rich_click/rich_click.py#L152-L173), but hit the wall I was expecting, not sure how click handles this

In particular [this stuff](https://github.com/seanbreckenridge/rich-click/blob/6aa3d0b2ccbb5535827c60ac4a9f6de02aa027e6/src/rich_click/rich_click.py#L214-L217):

```
    ## hmm -- show_default returns False even when ctx has a value set, since
    # show_default=False is the default for the parameter? Not sure how click
    # properly resolves this
```

OriBenHur-akeyless:
Are there any plans to fix this bug?

ewels:
[...]

PR #113, opened 2023-06-19
Add matrix strategy for CI + fix for Click 7.x.

### LINKED ISSUE #112: Regression in support for click 7.* in main branch (1.7.0dev, unreleased)
Was adding a matrix strategy to the CI and noticed the following:

```
git checkout main
pip install -U "click<8"
python examples/01_simple.py
```

Result: No output at all, and exit code 0.

I bring this up because the `setup.py` says that `click>=7` is supported.

It's possible I can get around to fixing this as part of my CI changes. There is a hint as to at least one thing that may be going on with the pytest results.

dwreeves:
Fixes #112, and lays groundwork for resolving #18.

# Overview

I was working on resolving #18 when I noticed that the solution to this involves branching logic for Click 8 and Click 7.

I also noticed this library is getting more popular (congrats!) and people are starting to use it for more things.

In light of all this, I thought that before I address #18, I should significantly expand the scope of the CI so that all versions of Click that you say are supported, actually are supported.

# Uncovering some regressions with Click 7.x support

In doing this, I discovered that the latest version of the code base in `main` actually has quite a few Click 7.x regressions.

In this code, I've fixed every regression I could see, except support for `rich_config`, which I could not get working.

Some of the regression fixes aren't the prettiest, but the fixes appear to work just fine.

# Where to go from here

I feel like you can either keep as is, or you can add deprecation warnings for support to click 7.x and then remove that support in a future release.


ewels:
This is massive, thanks for the work you've put into this!

I'm very happy to drop support for Click 7x. Click 8 has been around for a while now (since 2020) which I think is a fair period. And people can always install earlier versions of rich-click if that support was critical.

I'll merge now but if you fancy adding in some depreciation warnings within another PR that would be fab 👍🏻

dwreeves:
I'll add some deprecation warnings, that sounds fine to me. So I have two PRs on my plate:

- Deprecation warnings for Click 7.x
- Fixing #18.

Thank you for merging this!

ewels:
Perfect, thank you!

PR #117, opened 2023-07-17
Deprecate click 7.x support

### LINKED ISSUE #116: Deprecate Click 7.x
# Overview

https://github.com/ewels/rich-click/pull/113#issuecomment-1637160817

> I'm very happy to drop support for Click 7x. Click 8 has been around for a while now (since 2020) which I think is a fair period. And people can always install earlier versions of rich-click if that support was critical.
> 
> I'll merge now but if you fancy adding in some depreciation warnings within another PR that would be fab 👍🏻

# Acceptance Criteria

- Add deprecation warnings for users on Click 7.x. I will probably just add it in the `_click_compat.py` file and have the warning happen in the module scope, so it will run on the first import to that file.
- ~~Make sure there are `# TODO: remove when click 7 is deprecated` comments where Click 7.x compatibility was added.~~

dwreeves:
Resolves #116

Mostly self-explanatory PR.

I did not add TODO comments because all the parts of the code that need to be changed are easy enough to find in the code (basically, anywhere that `CLICK_IS_BEFORE_VERSION_8X` is referenced).

Lastly, on the topic of deprecation warnings, I removed `distutiils.version.LooseVersion` and swapped with `packaging.version.parse` since `distutiils` is deprecated and is slated to be removed in Python 3.12, via PEP-632 https://peps.python.org/pep-0632/ And I don't see any good reason for the next version release to not work with Python 3.12, even if only temporarily.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,6 +6,7 @@ setup(
         "rich>=10.7.0",
         "importlib-metadata; python_version < '3.8'",
         "typing_extensions",
+        "packaging",
     ],
     extras_require={
         "dev": ["pre-commit", "pytest", "flake8", "flake8-docstrings", "pytest-cov"],
--- a/src/rich_click/_compat_click.py
+++ b/src/rich_click/_compat_click.py
@@ -1,5 +1,17 @@
-from distutils.version import LooseVersion
+import warnings
 
 import click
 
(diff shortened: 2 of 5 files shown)


ewels (approved):
LGTM, thanks!

ewels:
Probably should have added a changelog entry for this.. 🤔

PR #122, opened 2023-10-04
fix issue where ctx.exit() did not emit the proper status code

### LINKED ISSUE #114: Exit codes aren't being propogated.
To reproduce run the following snippet of code:

```py
# foo.py
import rich_click as click

@click.command()
@click.pass_context
def foo(ctx: click.Context) -> None:
    print('exiting')
    ctx.exit(2)
    print('exited')

foo()
```

Invoking as `python foo.py || echo $?` prints just "exiting", which means the command is not failing.
After replacing the import with `import click`, `python foo.py || echo $?` correctly prints "exiting" followed by "2".

```
click                     8.1.3
rich                      13.4.2
rich-click                1.6.1
```

dwreeves:
- Resolve #114
- Added tests that replicate all of Click's behaviors in regards to exit codes.
- Created `RichMixin` class to reduce repetition of code.

I could not figure out a way to have full parity with click's functionality without re-implementing all of `click.BaseCommand.main()`. The reasons are explained in the comments for `click.BaseCommand.main()`, but tldr: . The downside to this is that it is not necessarily guaranteed to be forward compatible. But click is a relatively stable library with a slow and mature release cycle, so this is not a high cost.

I'd lik @ewels to review my use of a mixin. I am not sure if this was something you were avoiding for any particular reason. If you'd like I can avoid using the mixin approach, just let me know. I found it to be easier to switch to this so we don't need to reimplement all of that `main()` logic in multiple places.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ This PR closes a number of issues:
 - [#90](https://github.com/ewels/rich-click/issues/90): `click.ClickException` should output to `stderr`
 - [#88](https://github.com/ewels/rich-click/issues/88): Rich Click breaks contract of Click's `format_help` and its callers
 - [#18](https://github.com/ewels/rich-click/issues/18): Options inherited from context settings aren't applied
+- [#114](https://github.com/ewels/rich-click/issues/114): `ctx.exit(exit_code)` not showing nonzero exit codes.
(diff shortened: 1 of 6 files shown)


dwreeves:
I am moving this to #123.

PR #123, opened 2023-10-04
support more click classes

### LINKED ISSUE #38: Is `click.MultiCommand` supported?
Here's a minimal example using `click.MultiCommand` for lazy loading of subcommands. Is this already supported and I just haven't figured out how to use it, or is this not supported yet?

```py
import rich_click as click

class Cli(click.MultiCommand):
    def list_commands(self, ctx):
        return ["foo"]

    def get_command(self, ctx, name):
        if name == "foo":
            from foo import bar
            return bar
        raise NotImplementedError(f"The command '{name}' is not implemented.")

@click.group(cls=Cli)
def cli():
    pass
```

dwreeves:
- Resolves #38 and #114.
- Added tests that replicate all of Click's behaviors in regards to exit codes.
- Created two more classes: `RichBaseCommand` and `RichMultiCommand`. `RichBaseCommand` is used as a mixin.

This PR is based on #122, which I have closed. I am addressing both issues at once. A note from the other PR:

> I could not figure out a way to have full parity with click's functionality without re-implementing all of `click.BaseCommand.main()`. The reasons are explained in the comments for `click.BaseCommand.main()`, but tldr: . The downside to this is that it is not necessarily guaranteed to be forward compatible. But click is a relatively stable library with a slow and mature release cycle, so this is not a high cost.


dwreeves:
I am strongly considering combining this and #122 into a single PR. They are very closely related.

PR #129, opened 2023-10-07
update README-- push hello world example to the top and add branding

dwreeves:


README.md
```diff
@@ -18,7 +29,35 @@ click, formatted with rich, with minimal customisation required.
...
+
+### Simple Example
+
+To use rich-click in your code, replace `import click` with `import rich_click as click` in your existing click CLI:
+
+```python
+import rich_click as click
+
+@click.command()
+@click.rich_config(help_config=click.RichHelpConfiguration(use_markdown=True, width=60))
```
ewels, line 42: For a "simple example", I wonder if we should go _as simple as possible_ - maybe remove this line? Show it as a true 1:1 drop-in replacement.

Otherwise seeing `RichHelpConfiguration` here I think I would assume that this config is always required.

dwreeves: Got it. That is a very good point. I will just use the direct example from the click docs without anything else then.

dwreeves: I just pushed directly to `main` your requested change, plus new font for the logo. Thanks!

PR #132, opened 2023-10-08
two fixes

dwreeves:
Two changes right before the 1.7 release:

1. `__rich_context_settings__` is not fully overridden by `rich_config()`. I figure it should be possible for a user to define command-specific settings without overriding an existing `rich_console` should one exist. We were already doing that in the case of a `RichCommand` object, but this should also be true for callbacks.
2. `RichGroup.group` did not work properly for Click 7 because it does not take advantage of the `group_class` variable prior to Click 8. Slight oversight. 😅

### DIFF
--- a/src/rich_click/decorators.py
+++ b/src/rich_click/decorators.py
@@ -184,7 +184,10 @@ def rich_config(
         if isinstance(obj, (RichCommand, RichGroup)):
             obj.context_settings.update(extra)
         elif callable(obj) and not isinstance(obj, (Command, Group)):
-            setattr(obj, "__rich_context_settings__", extra)
+            if hasattr(obj, "__rich_context_settings__"):
+                obj.__rich_context_settings__.update(extra)
+            else:
+                setattr(obj, "__rich_context_settings__", extra)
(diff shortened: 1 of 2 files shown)


dwreeves:
Manually testing:

```python
# foo.py
from click import __version__
import rich_click as click

assert __version__.startswith("7")


@click.group("a")
def a():
    pass


@a.group("b")
def b():
    pass

@b.group("c")
def c():
    pass


a()
```

Ran this:

```
>>> python foo.py b c --help
                                                                                                                                                                             
 Usage: foo.py b c [OPTIONS] COMMAND [ARGS]...                                                                                                                               
                                                                                                                                                                             
╭─ Options ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --help      Show this message and exit.                                                                                                                                   │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

I can also confirm that when commenting out the code change I just added, it no longer works. 😬

PR #136, opened 2023-10-09
fix rich style bugs

### LINKED ISSUE #135: Bug: commands are always `bold cyan` and a few more style bugs.
Context: I was preparing to draft API documentation.

I was thinking a good way to do API documentation would be to strip all the formatting and only color one thing at a time (e.g. with red) so that the API documentation could clearly show one format at a time.

(Another more complex, but potentially interesting path, for API documentation would be to create an interactive widget along these same principles. 👀)

In any case, here is the code I was messing around with:

```python
import rich_click as click

click.rich_click.STYLE_OPTION = ""
click.rich_click.STYLE_ARGUMENT = ""
click.rich_click.STYLE_COMMAND = ""
click.rich_click.STYLE_SWITCH = ""
click.rich_click.STYLE_METAVAR = ""
click.rich_click.STYLE_METAVAR_APPEND = ""
click.rich_click.STYLE_METAVAR_SEPARATOR = ""
click.rich_click.STYLE_HEADER_TEXT = ""
click.rich_click.STYLE_FOOTER_TEXT = ""
click.rich_click.STYLE_USAGE = ""
click.rich_click.STYLE_USAGE_COMMAND = ""
click.rich_click.STYLE_DEPRECATED = ""
click.rich_click.STYLE_HELPTEXT_FIRST_LINE = ""
click.rich_click.STYLE_HELPTEXT = ""
click.rich_click.STYLE_OPTION_HELP = ""
click.rich_click.STYLE_OPTION_DEFAULT = ""
click.rich_click.STYLE_OPTION_ENVVAR = ""
click.rich_click.STYLE_REQUIRED_SHORT = ""
click.rich_click.STYLE_REQUIRED_LONG = ""
click.rich_click.STYLE_OPTIONS_PANEL_BORDER = ""
click.rich_click.ALIGN_OPTIONS_PANEL = "left"


@click.group()
@click.argument(
    "foo",
    type=click.FLOAT
)
@click.option(
    "--bar",
    default=3,
    show_default="someval",
    type=click.INT,
    help="Lorem ipsum",
    required=False)
def cli(foo: float, bar: int):
    """Help text for CLI"""
    pass


@cli.group("subgroup")
def subgroup():
    """Help text for subgroup"""


@cli.group("subcommand")
def subcommand():
    """Help text for subcommand"""


cli()
```

And I run it and, well...

image

Hmm. That doesn't seem right! I CTRL+F the code for `bold cyan` and find this:

image

And then I get digging and start seeing a few more issues:

- `STYLE_USAGE_COMMAND` only works as a global, not via the `rich_config`, and is only ever covered by code in `cli.main()`. This seems like it should not be the case.
- The `print usage` thing is not stylized at all.
- `"Try [blue]'{command} {option}'[/] for help."` also does not really allow for any formatting.

I think these should be possible to remedy before 1.7.0 release, so I will look into it. I will be honest, I have not paid too much close attention to the "rich" side of this code until now (been mostly focusing on the "click" side, plus drudgery like testing/linting etc.). So now is my time to dig a little deeper.

cc @ewels.

dwreeves:
Resolves #135. A handful of things were not hooked up properly.

Also, I added a new thing to the API, the `style_errors_suggestion_command`. I also included an example of that in the README.


dwreeves:
After studying the rich config API a little bit, and trying to grasp the intentions of each attribute, I am pretty comfortable with the way I approached this! I have also tested this change a decent bit, and also documented the one thing I am adding to the interface.

I will just merge without approval once the CI passes.

PR #138, opened 2023-10-09
Fix style bugs redux

dwreeves:
😅


dwreeves:
I added #137 as part of the coverage for this. If you ctrl+shift+F the `tests/` folder for "`for help`", nothing comes up. It would be useful to cover this.

PR #144, opened 2023-11-26
config refactors

dwreeves:
- Some internal refactors:
  - `rich_click.py` is exclusively the global config; all formatting has been moved to `rich_markup.py`.
  - `RichCommand` now makes use of methods in the super class: `format_usage`, `format_help_text`, `format_options`, and `format_epilog`.
  - Global `formatter` object has been removed from the code.
- Added `RichHelpConfiguration.build_from_globals()` classmethod, which pulls all configuration from `rich_click.py`.


dwreeves:
I decided to keep supporting Click 7.x because it's not a real burden at the moment to do so. 🤷 If that saves even just one person a minor headache I'll consider it worthwhile. But given the number of overrides that just reimplement Click 8.x features, it is admittedly starting to be a little silly inside the code, almost like we are vendoring click 8 for click 7 users.

PR #148, opened 2023-12-11
Add ruff + remove flake8 and isort

dwreeves:
- Added Ruff.
- Removed Flake8.
- Changed some docstrings as per the suggestion of Ruff.


dwreeves:
I love Ruff ❤️

ewels:
Nice! 👏🏻 Could use it to format the code instead of Black too, simplify the tool chain even more..

dwreeves:
@ewels I hadn't been keeping up with the Ruff news, I wasn't aware if the formatter was in a good place or not. If you think it is, then I'll go ahead with it!

Issue #134, opened 2023-10-08
RichHelpConfiguration changes for rich-click 1.8


dwreeves:
Brain dump on config stuff:

1. Decide on order of operations for resolving configs (e.g. env_var -> rich_config -> user defined default -> global default), and then document this order of operations.
    - It's entirely unclear to me whether env vars take precedence over rich_config, or vice versa. This could also itself be an option that the designers of CLI have control over. The reason it is unclear is because users may want to override config for compatibility and accessibility reasons, e.g. color blindness. See items 5-6 for more info.
2. Make it so environment variables do not resolve until the `RichHelpConfiguration` object is created, i.e. lazy load the environment.
3. Is it possible to create SSoT for defaults and type annotations when relying on both a global config and a config class? I think the answer is possibly no-- or alternatively that such a solution would be too metaprogrammatic to be developer friendly, so "no" with extra steps. But, we should at least have some sort of unit-test in place that enforces these things never fall out of sync. Something like:
    - `assert get_annotation(RichHelpConfiguration, config_option.lower()) is get_annotation(rich_click, config_option.upper())`.
    - `assert getattr(RichHelpConfiguration(), config_option.lower()) is getattr(rich_click, config_option.upper())`.
4. I think we should be making it so `rich_click.rich_click` is **only** for global config stuff, and move out all the parsing stuff. There are a few benefits to having a file devoted to global configuration:
    - It avoids conceptual errors in the code: none of the help formatting stuff should ever rely on global config options. By isolating the global config in a separate file and then never importing the global config into the file that parses the help stuff, we can easily avoid conceptual errors.
    - One file should focus on one thing; right now that file does two things.
    - This makes it so the entirety of `rich_click.rich_click` is itself a public interface, rather than just some of it.
    - `rich_click.rich_cick` module is far too entrenched and has too much a public-facing API to make changing the namespace on users worthwhile. Also, `import rich_click as click; click.rich_click.SOME_OPTION = ...;` is a perfectly reasonable API. What I am saying is, if you want to separate out the file into two different files, the only choice is to move the formatting stuff rather than moving the global config stuff.
5. Make it so each (or some subset?) config option can be set with `RICH_CLICK_*`
    - I am also thinking you'd want to deprecate all the other env vars as well. So _only_ `RICH_CLICK_*` is a valid choice. The obvious exception being the "force terminal" stuff that hooks into systems like Github Actions.
[...]

PR #149, opened 2023-12-11
[Draft] Use dicts to write configs

dwreeves:
Main changes:

- Add support for dicts to be passed into `@rich_config(help_config=...)`.
    - The current behavior, which I think I'm going to keep (but need to document), is to merge the dict into the parent context's help config.
    - This still leaves a lot of questions unanswered about the config resolution but I am overall satisfied with this API decision, since the typical user won't notice a difference between merge and instantiating a new object, whereas users looking for more customization haven't had a nice way to access merging. And it just seems to make sense that it would do this behavior: you set a large `RichHelpConfiguration()` at the top, and each sub-command may have a tiny tweak. That makes sense!

Misc. changes:

- Make typing more "Rich". At the end of the day the command class's type is `Type[RichCommand]`, and this typing is covariant, so it cannot ever return a super-type. The should can be safely updated to return that.
- Deprecate `RichCommand.help_config`. It didn't feel like this made sense back in 1.7, now it feels less like this makes sense, since rich help configs should be resolved in the context level (since they need to inherit from their parents).
- Add `stacklevel=2`s to the deprecation warnings.


dwreeves:
I'm likely going to keep this open for a while. I want to get down a sensible and flexible system for resolving the config, aimed at:

1. Zero difference to the vast majority of users who just want pretty outputs without much effort.
2. Lots of convenience, sensibility, zero astonishment for users looking for lots of customization.

And I'd rather do too few changes and too few API guarantees than too many.

Also, see #134.

dwreeves:
I think I am starting to settle on this configuration logic. I think it makes sense and is the least bad / least "astonishing" of all other alternatives. https://github.com/ewels/rich-click/issues/134#issuecomment-1872386532

PR #156, opened 2024-02-03
TypedDicts for option and command groups

dwreeves:
The interface for `command_groups` and `option_groups` is a little opaque. One way to help a little bit is to add `TypedDict` support, which is what this PR does.


dwreeves:
Oh woops, I forgot to run `pre-commit run -a`. Guess this will be a decent amount of work!

Issue #124, opened 2023-10-05
Cover `cli.py` in tests


dwreeves:
Caught a typo that I made (never got merged to main) that wasn't caught by tests in `rich_click.cli.patch`.

It should have been caught by mypy, too. (The pre-commit mypy run does not really do as much as it should, since it does not rely on type-checking using any imported packages.) But making the mypy checks more robust is a separate issue from this.

dwreeves:
Resolved via #158

PR #158, opened 2024-02-04
[Draft] Rewrite `rich-click` CLI with a `RichCommand`

### LINKED ISSUE #155: Use a RichCommand for the actual rich-click CLI tool + add formatting options
Right now the `cli.py` uses a custom CLI implementation, I believe for historic reasons. I _think_ this can be refactored into being a `RichCommand`.

There are two main benefits of this.

First: simpler, more idiomatic, and more testable code. Testability is the most important sub-bullet here. Right now `cli.py` is not covered by tests. See #124. This lack of test coverage of `cli.py` has caused multiple regressions over the last couple months; see #142 and #152.

Second: it would be to allow for things like passing of configuration flags, which can allow users who are aliasing `rich-click` to set options such as color themes. For example:

```bash
alias foo="rich-click --rich-config '{"\style_option\": \"red\", \"width\": 100}' foo"
foo bar
```

In the above example, the user sets some command in their Python environment named `foo` to an alias that invokes `rich-click` with custom style options.

There may be other options and flags that are of interest here; that's just the first that comes to mind.

---

This is doable and theoretically simple since Click allows for arbitrary passing of args to a command:

```python
# foo.py
import click

@click.command(name='my-cmd', context_settings=dict(
    ignore_unknown_options=True,
    allow_extra_args=True,
))
@click.option('--option1')
@click.option('--option2')
@click.pass_context
def cli(ctx, option1, option2):
    print(ctx.args)

cli()
```

```shell
>>> python foo.py --option1 a --option2 b c d e --f g
['c', 'd', 'e', '--f', 'g']
```

dwreeves:
Resolves #155 and adds a new feature to the `rich-click` CLI: `--rich-config [JSON]`, which allows the user to set a config for a command they're wrapping with `rich-click`. Note that `--rich-config` should also work not just for normal click commands but also rich-click commands.

This is a work in progress, and needs a lot of testing.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,7 @@ exclude = [
     "venv",
     "build",
     "sdist",
-    "tests/**",
+    "tests/fixtures/**",
     "examples/**",
 ]
 ignore = [
--- a/src/rich_click/cli.py
+++ b/src/rich_click/cli.py
@@ -157,10 +157,6 @@ def main(
     else:
         raise click.ClickException(f"No such script: {script_and_args[0]}")
 
-    if len(args) > 1:
-        if args[0] == "--":
-            del args[0]
-
     prog = module_path.split(".", 1)[0]
 
     sys.argv = [prog, *args]
--- a/src/rich_click/cli.py
+++ b/src/rich_click/cli.py
@@ -13,7 +13,6 @@ except ImportError:
     import importlib_metadata as metadata  # type: ignore[no-redef,import-not-found,unused-ignore]
 
 import click
-from rich.console import Console
 
 from rich_click.decorators import command as rich_command
 from rich_click.decorators import group as rich_group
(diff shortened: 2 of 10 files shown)


dwreeves:
Also addressing #124 in this PR. `cli.py` is now covered by tests 😄

dwreeves:
I think this should be fine.

One minor change. I removed this:

```python
    if len(args) > 1:
        if args[1] == "--":
            del args[1]
```

If someone is doing something _really_ weird, e.g. `rich-click -- -- flask -- run`, then this could break backwards compatibility? However, the thing is, Click already handles e.g. `rich-click -- flask -- run` just fine. And my unit-tests assert as much. So I believe that the manual removal of the `--` was always unnecessary.

I am making a note here in the very off chance this does break someone's setup, but I do not think that will be an issue.

PR #161, opened 2024-02-17
[Draft] Adding docs

dwreeves:


dwreeves:
This PR combines both some necessary code changes as well as very preliminary docs.

I need to get 1.8 in by the time [Astronomer Cosmos](https://www.astronomer.io/cosmos/) releases 1.4, as I am writing a blog post that will do some very minor cross promotion of Rich-Click (basically, I encourage users to alias their `dbt` command to `.venv_dbt/bin/rich-click dbt`). And without 1.8's performance improvements, I don't feel comfortable suggesting to users to do that, since the `dbt` command already has massive overhead and I want to add to that as little as possible.

I'm a busy guy though, so this may mean foregoing some of the docs stuff. I hope not but we'll see.

dwreeves:
Bad PR hygiene (should have been a separate PR), but the behaviors for patching in the `rich-click` CLI are cemented, properly unit-tested, etc. 😄

dwreeves:
> I need to get 1.8 in by the time [Astronomer Cosmos](https://www.astronomer.io/cosmos/) releases 1.4

I don't think this is happening (I'm just way too busy right now), but it's fine.

dwreeves:
This goes hard.

Just need to fix up a few things, and add the auto-code gen part.

image

dwreeves:
Sooooooooooooooooooo close. So so so so very close.

Issue #164, opened 2024-03-10
Expose a way to send usage information to stderr


Julian:
I was quickly evaluating switching/using this [on a project of mine](https://github.com/bowtie-json-schema/bowtie), which looks quite nice (originally I found it simply looking for a good command-group plugin for click, but given I already use rich this seems like a nice fit).

I have one main issue though from trying to drop this in -- my tests fail because usage errors seem to go to stdout rather than stderr.

Specifically if I run `bowtie non-existing-subcommand` after simply changing to `import rich_click as click` I get usage output on stdout.

I see the README mentions that the rich console I can modify is *separate* from the one used internally. It seems the internal one is the one that needs modifying to change this behavior. Looking at the code, I see there's a [create_console](https://github.com/ewels/rich-click/blob/main/src/rich_click/rich_help_formatter.py#L20) that takes a `file` where I'd love to pass `sys.stderr`, but as far as I can quickly tell there's nothing that calls that, and/or no way I can actually pass that argument through via `rich.click_config()`. Alternatively I don't quite see a way to inject a `RichHelpFormatter`.

Apologies if this exists and I've missed it. I'm happy to send a PR if need be, though to do so I'd love to know whether the current behavior (sending to stdout) is intentional / whether you want to keep it, and if so but you still would support some way of changing where it goes, whether the intention is to expose `create_console` somewhere or else to only expose `file` directly via `click_config`.

Thanks!

dwreeves:
Hi @Julian, thank you for opening this issue!

> but as far as I can quickly tell there's nothing that calls that

Yep, that's a bug that I noticed and will be fixed in 1.8.0. I am already working on that.

> I have one main issue though from trying to drop this in -- my tests fail because usage errors seem to go to stdout rather than stderr.

One of rich-click's objectives is to be a seamless drop-in that mimics click. In this case, there is a disparity between how rich-click emulates click, and is thus a bug.

I think I/we can fix this as part of a patch release to 1.7.x.

If you want to do a PR based on the "1.7.x" branch, I can merge it in and release it!

If you'd rather me work on it, I won't be able to get to this until next weekend.

Julian:
Thanks for the response! Very glad to hear.

I'll give it a quick look.

> One of rich-click's objectives is to be a seamless drop-in that mimics click. In this case, there is a disparity between how rich-click emulates click, and is thus a bug.

Julian:
Thanks for the response! Very glad to hear.

I'll give it a quick look.

[...]

PR #166, opened 2024-03-11
Send usage information to stderr when something fails.

Julian:
Matches click's behavior.

Closes: #164

### DIFF
--- a/src/rich_click/rich_help_formatter.py
+++ b/src/rich_click/rich_help_formatter.py
@@ -1,3 +1,4 @@
+import sys
 from typing import Any, IO, Optional
 
 import click
@@ -83,7 +84,7 @@ class RichHelpFormatter(click.HelpFormatter):
             max_width = config.max_width or max_width
         super().__init__(indent_increment, width, max_width, *args, **kwargs)
         self._config = config or get_module_config()
-        self._console = create_console(self._config)
+        self._console = create_console(self._config, file=sys.stderr)
 
     @property
     def config(self) -> RichHelpConfiguration:


dwreeves:
I'll admit I am a little confused about something with your original report and this PR, I'm hoping there is a way we can resolve it.

I have a file called `foo.py` with the following contents:

```python
# import rich_click as click
import click


@click.group("my-command")
@click.argument("foo")
@click.option("--bar", "-b", help="Lorem ipsum", show_default="someval")
def cli(foo, bar):
    """Help text for CLI"""

@cli.command("subcommand")
def subcommand(foo, bar):
    """Help text for subcommand"""

if __name__ == "__main__":
    cli()
```

And I use this script to execute `foo.py` and examine the outputs:

```python
from subprocess import Popen, PIPE
process = Popen(["python", "foo.py", "--help"], stdout=PIPE, stderr=PIPE)
stdout, stderr = process.communicate()
print(f"{len(stdout)=}", f"{len(stderr)=}")
```

When I run it with `import click`, I get the following: `len(stdout)=225 len(stderr)=0`

When I run with `import rich_click as click`, I get the following: `len(stdout)=3447 len(stderr)=0`

When I run with `import rich_click as click`, and also use your code change, I get the following: `len(stdout)=1 len(stderr)=3446`

Which is to say, the default behavior of writing to `stdout` seems correct and in line with what Click does, but this PR seems to change it so the default behavior is that it writes to stderr.

---

What I thought you had in mind when you said we aren't printing to stderr was more something like this. The following code adds a `--fake-flag` to trigger an error:

```python
from subprocess import Popen, PIPE
process = Popen(["python", "foo.py", "--fake-flag"], stdout=PIPE, stderr=PIPE)
stdout, stderr = process.communicate()
print(f"{len(stdout)=}", f"{len(stderr)=}")
```

When I run it with `import click`, I get the following: `len(stdout)=0 len(stderr)=117`

When I run with `import rich_click as click`, I get the following: `len(stdout)=2074 len(stderr)=0`

^ This to me is erroneous behavior on rich-click's part. By default, rich-click _should_ write to `stdout`, it's just in error-handling situations that it should be writing to `stderr`, but it isn't.

Do you agree with everything I've stated? Let me know if anything here looks wrong.

---

In 1.8.0, which I can potentially release an alpha version of soon (mostly, I'm just waiting to [finish the docs](https://github.com/ewels/rich-click/pull/161), aaah), you'll notice the `__init__` for `RichHelpFormatter` looks different than in `1.7.x`:

```python
    def __init__(
        self,
        indent_increment: int = 2,
        width: Optional[int] = None,
        max_width: Optional[int] = None,
        *args: Any,
        console: Optional["Console"] = None,
        config: Optional[RichHelpConfiguration] = None,
        file: Optional[IO[str]] = None,
        **kwargs: Any,
    ) -> None:
```

TLDR, 1.7.0 was a massive shake-up in rich-click's internals, but some of the internals were not fully fleshed out, and so a big part of what 1.8.0 is doing is making those internals better.

---

I will say, in the dev branch for `1.8.0a`, we are currently _not_ writing to `stderr` for error messages. So the erroneous behavior you've flagged is still existent in `1.8.0`.

This is something we could potentially address in `1.8.0`, and it may be difficult to address in `1.7.x`. And in that case, I'm really sorry I encouraged you to push a change to `1.7.x` 😦.

---

TLDR of everything:

- Based on what I can tell, `stderr` should be written to _only_ for error messages
- rich-click does not currently do that, hence that is a bug.
- It is possible but may be difficult to coherently implement this change in `1.7.x`, however `1.8.0`'s internals have been reworked a bit and it should be a little easier there.
- Again, really sorry I encouraged you to push changes to `1.7.x`; I was typing that from work and didn't have time to fully investigate everything relating to the matter.

Julian:
> Do you agree with everything I've stated? Let me know if anything here looks wrong.

Sorry, it's me who was clearly quite lazy and didn't look carefully after getting my own suite to pass! I indeed only have test cases for the error behavior and didn't even check the "normal" one.

Everything you wrote looks correct to me!

> This is something we could potentially address in 1.8.0, and it may be difficult to address in 1.7.x. And in that case, I'm really sorry I encouraged you to push a change to 1.7.x 😦.

Don't worry about it, clearly I didn't spend long for better or worse, I mostly gave up after I couldn't get the epilog change fixed. I'll try to file a separete issue for that one. But I'm more than happy either to wait for 1.8, or else to give this another shot when I have more spare time than I do at the minute, if you haven't gotten to it yourself (right now is super busy as I'm busy with Google Summer of Code, but hopefully once that gets started properly I may have more time...)

PR #185, opened 2024-04-21
Docs cleanup

dwreeves:
- Primarily I am cleaning up the docs as per our email discussions
- I also did a few changes to the CI while I was at it:
  - Using `uv` where I can. (It's a lot faster.)
  - Added `actionlint` to pre-commit.


dwreeves:
Docs are almost done, or done, depending on the importance of the final feature.

See docs here:

https://dwreeves.github.io/rich-click/

The only thing missing, i.e. the final feature, is buttons on the editor that correspond with styles, and will set everything to it once clicked:

- Default
- Blank
- Example 1
- Example 2

PR #209, opened 2024-11-12
Add blog post about styles and patch release for 1.8.4

dwreeves:
- Wrote a blog post about styling ideas!
- Patch release `1.8.4` - Minor fix that came up when I was writing the blog post (I wanted to add the `Gradient()` in the header text, but only `str` was supported.)


dwreeves:
I think it should be fine to just merge. I want this ready and also promoted before your PyCon session!

PR #215, opened 2024-11-24
fix/exception-in-console

sankarngrjn:
Previously, when passing rich_console in context_settings, all outputs except exceptions were rendered under the provided console. This caused inconsistencies in the output format, as exceptions would bypass the rich console, leading to a mismatch in styling or behavior.

This change ensures that exceptions are also captured and displayed within the provided rich_console. By addressing this, all outputs, including exceptions, will now conform to the same console formatting, maintaining consistency across the board.

Ref. Past

Ref. Present

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Version 1.8.5 (2024-12-01)
+
+- Fix issue with error messages not using user-defined styles. [[#215](https://github.com/ewels/rich-click/pull/215)] ([@sankarngrjn](https://github.com/sankarngrjn))
+
 ## Version 1.8.4 (2024-11-12)
 
 - Support `rich.text.Text()` objects for `header_text`, `footer_text`, `errors_suggestion`, and `errors_epilogue`.
--- a/src/rich_click/__init__.py
+++ b/src/rich_click/__init__.py
@@ -6,7 +6,7 @@ The intention is to provide attractive help output from Click, formatted with Ri
 customisation required.
 """
 
-__version__ = "1.8.4"
+__version__ = "1.8.5"
 
 # Import the entire click API here.
 # We need to manually import these instead of `from click import *` to force
--- a/src/rich_click/rich_command.py
+++ b/src/rich_click/rich_command.py
@@ -167,10 +167,10 @@ class RichCommand(click.Command):
                 if not standalone_mode:
                     raise
                 if ctx is not None:
-                    config = ctx.help_config
+                    formatter = ctx.make_formatter()
                 else:
                     config = self._generate_rich_help_config()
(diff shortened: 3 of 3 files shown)


tests/test_config.py
```diff
@@ -71,7 +71,7 @@ def test_global_config_equal_config_defaults() -> None:
 
 def test_config_from_globals_behavior() -> None:
     original_style_option_value = rc.STYLE_OPTION
-    rc.STYLE_OPTION = "new-value"
+    rc.STYLE_OPTION = "bold cyan"
```
dwreeves, line 74: The reason this is `new-value` and not `bold cyan` is because we want to assert a change occurred. Because `bold cyan` is the default, this change means the test is no longer testing the thing it should be.

dwreeves (changes requested):
Hello, thank you for your PR. Your description of the issue makes sense, although I have one note on the test changes.

tests/test_config.py
sankarngrjn: Thank you for pointing that out and guiding me in the right direction. I revisited my changes, ensured they align with the intended behavior, and have pushed an updated fix. Could you kindly review it and let me know if it works as expected?

dwreeves: Thanks for the speedy change! Left a comment in the main thread, tldr I'll either review manually myself end of week, or we can speed this up if you give a reproducible example. Your call.

sankarngrjn:
Hi, sorry for the delayed response—got caught up with some personal work!

I've also noticed a few other drawbacks and have pushed an update. I believe that whenever we're using rich_console in context settings, everything should consistently fall under the rich_console. It didn’t seem logical to only use the console in these scenarios when a rich context is present, so I’ve made this adjustment accordingly.


Here’s an example to demonstrate the change:
`
import rich_click as click

from catppuccin.extras.rich_ctp import mocha
from rich.console import Console
from rich.theme import Theme

mocha = mocha.styles
mocha["none"] = f" on {mocha['base']}"

console = Console(theme=Theme(mocha), width=80)

context_settings = dict(
    rich_console=console
)

# context_settings = None

@click.command(cls=click.RichCommand, context_settings=context_settings)
def greet():
    """Print a friendly greeting."""
    click.echo("Hello! Welcome to the RichClick example!")

if __name__ == "__main__":
    greet()
`

When the line context_settings = None is uncommented, the help error will behave as usual.

Regarding the test case, I’ll address that in a future update when we revamp the testing approach, as you mentioned. For now, I hope the example provided helps with the review.

Let me know your thoughts!

dwreeves (approved):
LGTM

dwreeves:
Very cool. Ran it and it works great! Thank you for your work on this @sankarngrjn, I will get it out shortly.

PR #217, opened 2024-12-05
WIP Add configuration option to use STDERR for output

alexanderilyin:
@ewels is adding `config.stderr` property similar to other settings is the right way to implement this?

### DIFF
--- a/src/rich_click/rich_help_formatter.py
+++ b/src/rich_click/rich_help_formatter.py
@@ -42,6 +42,7 @@ def create_console(config: RichHelpConfiguration, file: Optional[IO[str]] = None
                 "usage": config.style_usage,
             }
         ),
+        stderr=True,
         color_system=config.color_system,
         force_terminal=config.force_terminal,
         file=file,


ewels:
Thanks for this @alexanderilyin!

There's already been some discussion around printing to stderr, see #164. If I understand @dwreeves correctly, Click usually prints to stdout for help messages and stderr for error messages. Rich-click aims to emulate this behaviour.

That said, I'd be happy to make this behaviour configurable as long as the default matches Click. So in other words, add a config option as you suggest and make it opt-in.

Does that sound ok?

alexanderilyin:
@ewels sound like a plan. Yep I love that `rich-click` is Plug & Play but for my CLI app I want all human readable things go to STDERR for now, and STDOUT is reserved for programmatic interactions, in other words I don't want JSON in STDOUT be broken due to any other output, and I do need to control exit codes. Note to self: I can grab STDERR as is and internally merge it with JSON from STDOUT on "client" side.

Thanks in advance, I'll put PR sometime this week, cause I don't want to have a fork and all overhead related to it.

alexanderilyin:
@ewels one more opt-in feature I want to add is `help` for arguments. Wonder if any work was happening in that direction. I looked quickly around but seems that there is no support for that yet.

ewels:
Yes I've wondered about that in the past. Again it moves us further from the original `click` syntax so needs some care, but it may well be worth it. Please create an issue for that so we can track it and discuss that there.

dwreeves:
Help for arguments is something I'm thinking about for 1.9 or 2.0. I'm personally open to it.

dwreeves:
Hi @alexanderilyin,

Via #255, `help_to_stderr` can now be set like this:

```python
@click.command(context_settings={"help_to_stderr": True})
def cli() -> None:
    """CLI help text"""
```

After many internal refactors, it just doesn't really work to have it in help config settings.

Also, `@click.argument()` now accepts a help= kwarg.

This will all be available in 1.9.0, which should be out in a few weeks, let's say first week of September.

Issue #223, opened 2025-02-19
Failure with click 8.2.0 - TypeError: Parameter.make_metavar() missing 1 required positional argument: 'ctx'


s3v-:
Hi,

Click 2.8.0 recently entered in Debian [1], and the example provided in your main page fails:

```python
 Usage: 11_hello.py [OPTIONS]                                                                                                                                   
                                                                                                                                                             
 Simple program that greets NAME for a total of COUNT times.                                                                                                 
                                                                                                                                                             
Traceback (most recent call last):
  File "/build/11_hello.py", line 12, in <module>
    hello()
    ~~~~~^^
  File "/usr/lib/python3/dist-packages/click/core.py", line 1429, in __call__
    return self.main(*args, **kwargs)
           ~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/rich_click/rich_command.py", line 151, in main
    with self.make_context(prog_name, args, **extra) as ctx:
         ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/rich_click/rich_command.py", line 260, in make_context
    return super().make_context(info_name, args, parent, **extra)  # type: ignore[return-value]
           ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/click/core.py", line 1172, in make_context
    self.parse_args(ctx, args)
    ~~~~~~~~~~~~~~~^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/click/core.py", line 1183, in parse_args
    value, args = param.handle_parse_result(ctx, opts, args)
                  ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/click/core.py", line 2403, in handle_parse_result
    value = self.process_value(ctx, value)
  File "/usr/lib/python3/dist-packages/click/core.py", line 2342, in process_value
    value = self.callback(ctx, self, value)
  File "/usr/lib/python3/dist-packages/click/core.py", line 1025, in show_help
    echo(ctx.get_help(), color=ctx.color)
         ~~~~~~~~~~~~^^
  File "/usr/lib/python3/dist-packages/click/core.py", line 723, in get_help
    return self.command.get_help(self)
[...]

PR #224, opened 2025-02-19
Fix a couple of issues with click 8.2

ppentchev:
Hi,

First of all, thanks a lot for writing and maintaining rich-click!

I am the maintainer of the Debian package of click; I uploaded click 8.2.0 to the Debian unstable suite a couple of days ago, and it turns out that might have been a bit premature, since there are a couple of incompatibilities that affect several other Python libraries. As noted in #223, rich-click is one of those - the click authors changed both the `make_metavar()` method and the behavior of a command-line program when run without specifying a required subcommand.

What do you think about these two changes that try to fix these problems?

Thanks in advance for your time, and keep up the great work!

### DIFF
--- a/src/rich_click/_compat_click.py
+++ b/src/rich_click/_compat_click.py
@@ -5,6 +5,7 @@ try:
     click_version = click.__version__
 except Exception:
     # Click 9+ deprecated __version__, so all these checks must necessarily be False if __version__ doesn't exist.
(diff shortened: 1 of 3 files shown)


dwreeves:
Looks good, thank you so much for your work.

I'm not sure what the test failure is about.

I'll make sure to release this ASAP and credit you, I'll also get test coverage for this in the future.

PR #230, opened 2025-04-06
Properly override `click.HelpFormatter.getvalue`

ofek:
https://click.palletsprojects.com/en/stable/api/#click.HelpFormatter.getvalue

If the class attribute `export_console_as` on `RichContext` (or a subclass) is set to a value that is not `None` then getting the value of the buffer returns the raw text that was printed to the underlying console rather than returning an empty string from the parent class' method.

I'm a maintainer of https://github.com/mkdocs/mkdocs-click and I noticed that users of `rich-click` commands were not rendering the usage properly because [this line](https://github.com/mkdocs/mkdocs-click/blob/a614ab60b2f44ff5faa6d9c6554e6a857dacef5c/mkdocs_click/_docs.py#L196) was returning nothing. The choice to return text here is both out of necessity because the current context is inaccessible to that method and also because we happen to desire text so that rendering does not contain any formatting:

image

### DIFF
--- a/src/rich_click/rich_help_formatter.py
+++ b/src/rich_click/rich_help_formatter.py
@@ -140,3 +140,9 @@ class RichHelpFormatter(click.HelpFormatter):
     def write_abort(self) -> None:
         """Print richly formatted abort error."""
         self.console.print(self.config.aborted_text, style=self.config.style_aborted)
+
+    def getvalue(self) -> str:
+        if self.console.record:
+            return self.console.export_text()
+
+        return super().getvalue()
--- a/src/rich_click/rich_help_formatter.py
+++ b/src/rich_click/rich_help_formatter.py
@@ -143,6 +143,6 @@ class RichHelpFormatter(click.HelpFormatter):
 
     def getvalue(self) -> str:
         if self.console.record:
-            return self.console.export_text()
+            return self.console.export_text(clear=False)
 
         return super().getvalue()
--- a/src/rich_click/rich_help_formatter.py
+++ b/src/rich_click/rich_help_formatter.py
@@ -143,6 +143,6 @@ class RichHelpFormatter(click.HelpFormatter):
 
     def getvalue(self) -> str:
         if self.console.record:
-            return self.console.export_text(clear=False)
+            return self.console.export_text(clear=False, styles=True)
 
         return super().getvalue()


src/rich_click/rich_help_formatter.py
```diff
@@ -140,3 +140,9 @@ def write_usage(self, prog: str, args: str = "", prefix: Optional[str] = None) -
     def write_abort(self) -> None:
         """Print richly formatted abort error."""
         self.console.print(self.config.aborted_text, style=self.config.style_aborted)
+
+    def getvalue(self) -> str:
+        if self.console.record:
+            return self.console.export_text()
```
ofek, line 146:
```suggestion
            return self.console.export_text(clear=False)
```

ofek:
Good idea, thanks! I just pushed a commit that does that.

dwreeves:
Thanks.

Upon further thinking (refer to the discussion in the linked issue), there may be some additional work that needs to be thought through.

If we switch to something more compliant with the `HelpFormatter` API, as per the other discussion, that means `getvalue()` is called at the very end of the runtime context and prints all the output, meaning it would need to have the `styles=True` kwarg set as well for a typical runtime. However, this should probably be configurable, which could be done by adding a kwarg to `getvalue()` (altho that could be weird to do that though as getvalue() is a common duck-typed method name for interacting with IO-like objects and typically accepts no args), or more likely by adding a kwarg to `RichHelpFormatter()`.

We may also want to consider adding a more explicit `def export(as_: Literal["html", "svg", "text"] = "text", **export_kwargs: Any) -> str` method to the `RichHelpFormatter` as well, which could then be called at the end of the runtime context where `rich-click --output ...` is specified (instead of `getvalue()`).

Sorry, need to sit on this for a bit. I am supportive of where you / we are headed here, but I want to make sure it is good for general purpose usage. I do understand your concerns are a little more narrow and you want to support your users of `mkdocs-click`. I'm on board with that, users who simultaneously use rich-click and mkdocs-click are also my users as well, I just also have a lot more users than that.

src/rich_click/rich_help_formatter.py
```diff
@@ -140,3 +140,9 @@ def write_usage(self, prog: str, args: str = "", prefix: Optional[str] = None) -
     def write_abort(self) -> None:
         """Print richly formatted abort error."""
         self.console.print(self.config.aborted_text, style=self.config.style_aborted)
+
+    def getvalue(self) -> str:
+        if self.console.record:
+            return self.console.export_text(clear=False)
```
ofek, line 146:
```suggestion
            return self.console.export_text(clear=False, styles=True)
```

ofek:
I made the minimal requested change because I don't have time and to a lesser extent I don't want to mess up 🙂

ofek:
You're a very good maintainer btw, it's been nice interacting with you 🙇‍♂️

dwreeves:
Thanks @ofek! And sorry it took me a little bit of time to fully understand what is going on. I appreciate your contribution and I will take it from here.

Issue #48, opened 2022-03-21
rich markup shows when using sphinx-click


ri0t:
Currently, the click-man-page rendering and sphinx html docs get pretty messed up when using rich-click.
Here's an example of a man page output:

```
ISO(1)                                        iso Manual                                       ISO(1)

NAME
       iso - Main Isomer CLI

SYNOPSIS
       iso [OPTIONS] COMMAND [ARGS]...

DESCRIPTION
       [bold cyan on blue] :diamonds: Isomer[/][white on blue] Management Tool [/]

       This tool supports various operations to manage Isomer instances.

       [yellow]  :warning: Most of the commands are [u]grouped[/u].[/]

       To obtain more information about the groups' available subcommands/groups, try:

       [bright_cyan]  iso [/]

       To display details of a command or its subgroups, try:

       [bright_cyan]  iso   [..]  --help[/]

       To get a map of all available commands, try:

       [bright_cyan]  iso cmdmap[/]

OPTIONS
       -e, --env, --environment [blue|green|current|other]
              Override environment to act on (CAUTION!)
```
Notice the `--env` option - it is not immediately obvious if that is weird markup or actual option choices.

I think it would make sense to somehow throw all the markup out in manpages. With sphinx-generated html or similar things that could render them - not sure how to proceed.

ewels:
Hmm, I've not heard of [click-man-page](https://github.com/click-contrib/click-man) before, and I guess it's the [sphinx-click](https://github.com/click-contrib/sphinx-click) plugin? I'd need to look into how they are fetching the click output first - if you're able to do some detective work that would help.

However, hopefully https://github.com/ewels/rich-click/issues/19 will help. The aim there is to make it easier to return a rich renderable object instead of always printing. Then it would be fairly easy to get a plaintext copy of the help without any markup. If the above plugins are using regular click function calls to fetch the help strings, it might be that we can refactor the rich-click code to better emulate native click. Bit of a long shot, but may be possible.

dwreeves:
As far as rst is concerned, we have it on our roadmap to support rst docstrings in #172.

And HTML generation is supported via: `rich-click --output=html path.to.my:cli --help`

Rendering the CLI help text _as_ simple rst is a different story. We can potentially add this to the `--output=?` options in the `rich-click` CLI.

[...]

PR #231, opened 2025-04-09
[WIP] Update RichHelpFormatter to buffer and render later

dwreeves:
This PR closer aligns the `RichHelpFormatter` with how the base Click `HelpFormatter` works in regards to eager printing. This is in relation to issue #48 and a request from @ofek.


src/rich_click/rich_command.py
```diff
@@ -180,15 +182,9 @@ def main(
...
-
-                if self.context_class.errors_in_output_format:
-                    export_console_as = self.context_class.export_console_as
-                else:
-                    export_console_as = None
-
-                rich_format_error(e, formatter, export_console_as=export_console_as)
+                formatter = self._error_formatter()
+                formatter.write_error(e)
+                click.echo(formatter.getvalue(), file=sys.stderr)
```
ofek, line 187: Out of curiosity, what do you think about never using Click to output and relying exclusively on Rich? That's actually what I assumed was happening already. Click uses `colorama` on Windows which is basically unmaintained and has some odd side effects. I would love for Click to stop using that eventually and actually remove terminal capabilities by default (see this [closed issue](https://github.com/pallets/click/issues/2776)).

dwreeves: That's a good suggestion and I'll consider it. Rich does have its own legacy Windows color handling, so wrapping in `click.echo()` should not be necessary for that situation. I need to make sure the other small benefits of `echo()` are also being handled by Rich in how it renders the string generated by `console.print()`.

dwreeves: So, I don't want to do this personally. I'm open to PRs if anyone else wants it. #233

dwreeves: Basically, there are a handful of minor things I ran into that suggest to me this could be a very minorly annoying breaking change, and it's a lot of work to "fully" implement for something that doesn't impact the majority of both developers or even users. If there are fundamental issues with Click's use of Colorama, I would probably more likely bring it up with Click themselves. None of that is to say the feature wouldn't be welcome, of course, but I will not personally be prioritizing it over other things I am more interested in prioritizing.

ofek: Did you discover any missing functionality or edge case handling on Rich's side?

dwreeves: No, I discovered it changes my unit-tests from being colorless to having colors with the config we currently have. That is enough for me to believe the change can be classified as "minorly annoying breaking" (imagine a person's unit tests break the same ways ours did, in a cryptic way that they don't know how to make work in the old way, because of a change from click.echo to rich.print in a minor release; they would probably be rightly annoyed.) This would make it inappropriate for anything other than a major release if we make `rich.print()` the default, rather than as a config option that's turned off by default (which would be in-scope for a minor release).

ofek:
Do you know when this might be released?

dwreeves:
Oh neat, looks like a cool project. And yeah this is definitely a huge problem with `stdio` transit. I think the same general idea of:

```python
    try:
        import rich_click
    except ImportError:
        pass
    else:
        if isinstance(command, rich_click.RichCommand):
            ctx.console.record = True
            ctx.console.file = open(os.devnull, "w")
```

will work to prevent output to `stdout` while also being invisible to non-rich-click users. Otherwise, will just need to wait for 1.9.

PR #238, opened 2025-05-18
Use `importlib.metadata` to get click version

finsberg:
Get `click` version using `importlib.metadata`, which is also suggested by the `DeprecationWarning` that is emitted
```
  /home/shared/rich-click/src/rich_click/_compat_click.py:5: DeprecationWarning: The '__version__' attribute is deprecated and will be removed in Click 9.1. Use feature detection or 'importlib.metadata.version("click")' instead.
    click_version = click.__version__
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Version 1.8.8 (2025-03-09)
+
+- Make text wrap instead of using ellipses for overflowing metavars in options tables.
+- Added `--errors-in-output-format` flag to `rich-click` CLI.
+- Actually fixed regression in stderr handling [[#164](https://github.com/ewels/rich-click/issues/164)].
+
 ## Version 1.8.7 (2025-03-08)
 
 - Add ability to turn off option/command deduplication in groups [[#226](https://github.com/ewels/rich-click/issues/226)]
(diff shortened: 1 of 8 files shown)


dwreeves:
Hi @finsberg! Thanks for the contribution!

Two notes.

(1)

I would rather use `click.__version__` because it is more performant and doesn't require the extra dependency.

When choosing between two alternative approaches that do the same thing, we want the lighter weight and more performant solution because we want rich-click to add as little overhead as possible.

We note this as a performance improvement in a blog post: https://ewels.github.io/rich-click/blog/2024/04/30/version-1.8/

I understand Click 9.0 is deprecating `__version__`, unfortunately.

What I would prefer to do instead is just filter the warning; `warnings` is often (not always) already loaded into `sys.modules` and it's faster to do all of that. And when `__version__` does not exist, we can just assume a very new version of Click.

If `__version__` is being loaded only via the module level `__getattr__` then you can also do a hasattr check to avoid `import warnings` for pre-9.0, which is best of both worlds.

(2) This needs to be pointing not to the `main` branch but to the `1.8.x` branch so that we can suppress the warning in older versions of rich-click. `main` currently points to a `1.9.0` development version that won't be out for a while.

finsberg:
@dwreeves Thanks for clarifying, and sorry for missing the point about performance. Trying to add a context manager to suppress the warning (and also change the target branch to `1.8.x`). Will open a new PR

PR #239, opened 2025-05-19
Supress depreaction warning from click when accessing __version__

finsberg:
See https://github.com/ewels/rich-click/pull/238 for more context.

### DIFF
--- a/src/rich_click/_compat_click.py
+++ b/src/rich_click/_compat_click.py
@@ -1,8 +1,12 @@
+import warnings
+
 import click
 
 
 try:
-    click_version = click.__version__
+    with warnings.catch_warnings(module="click", category=DeprecationWarning, action="ignore"):
+        # Ignore deprecation warnings for click 8.0
+        click_version = click.__version__
 except Exception:
     # Click 9+ deprecated __version__, so all these checks must necessarily be False if __version__ doesn't exist.
     CLICK_IS_BEFORE_VERSION_82 = False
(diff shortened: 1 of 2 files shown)


dwreeves:
Thanks, although there is an issue. We support as far back as Python 3.7 for rich-click 1.8 (soon we will drop 3.7 and go to 3.8+). The syntax you are using is only compatible with 3.11+. You would need to use a different syntax.

src/rich_click/_compat_click.py
```diff
@@ -1,8 +1,12 @@
+import warnings
+
 import click
 
 
 try:
-    click_version = click.__version__
+    with warnings.catch_warnings(module="click", category=DeprecationWarning, action="ignore"):
+        # Ignore deprecation warnings for click 8.0
+        click_version = click.__version__
```
dwreeves, line 9:
```suggestion
    with warnings.catch_warnings(module=click):
        warnings.simplefilter(category=DeprecationWarning, action="ignore")
        click_version = click.__version__
```

dwreeves:
Another issue you may run into when testing click 8.2.0 is the pre-commit checks will fail on all files because of new typing changes. If you make the change I suggest above, I will override any failed CI and push those changes and I'll handle the typing changes.

dwreeves:
Hello @finsberg,

Thanks for the changes! I will take on the mypy stuff. I'll have this out in a couple hours.

finsberg:
> Hello @finsberg,
> 
> Thanks for the changes! I will take on the mypy stuff. I'll have this out in a couple hours.

Hi @dwreeves, I already fixed the mypy stuff here: https://github.com/ewels/rich-click/pull/240 :)

Issue #241, opened 2025-05-19
Click 8.2 no_args_is_help behavior


pdimens:
Hi, longtime fan of your work here.

The new Click v.8.2 changes the behavior of the `no_args_is_help` option to return an exit code of `1`. This is generally fine, but it doesn't jive well with `rich-click`, as the latter outputs an empty error message when using the `no_args_is_help` option:

MWE Click 8.2
```python
import click

@click.command(no_args_is_help=True)
@click.argument("text")
def cli(text):
    '''Here is text'''
    print(text)

if __name__ =='__main__':
    cli()
```
which, when run without arguments, outputs:
```bash
➜ ./clicktest.py
Usage: clicktest.py [OPTIONS] TEXT

  Here is text

Options:
  --help  Show this message and exit.
```
Notably, there is no extra error text, despite the return code of `1`

## rich-click behavior
However, replacing `import click` with `import rich-click as click` results in the empty error message, presumably because Click now gives a return code of `1`. 
```python
#! /usr/bin/env python
import rich_click as click

@click.command(no_args_is_help=True)
@click.argument("text")
def cli(text):
    '''Here is text'''
    print(text)

if __name__ =='__main__':
    cli()
```

```bash
➜ ./clicktest.py     
                                                            
 Usage: clicktest.py [OPTIONS] TEXT                         
                                                            
 Here is text                                               
                                                            
╭─ Options ────────────────────────────────────────────────╮
│ --help      Show this message and exit.                  │
╰──────────────────────────────────────────────────────────╯
                                                            
 Usage: clicktest.py [OPTIONS] TEXT                         
                                                            
 Try 'clicktest.py --help' for help                         
╭─ Error ──────────────────────────────────────────────────╮
│                                                          │
╰──────────────────────────────────────────────────────────╯
```

dwreeves:
👀 Working on some fixes right now. I'll make sure this is fixed, and I'll add it to the coverage as well.

dwreeves:
Oh one more note. `python clicktest.py` will return an exit code of 2 now. That's the exit code returned by base Click in this situation.

PR #242, opened 2025-05-19
Updates for click 8.2

dwreeves:


dwreeves:
Added fix for #241

dwreeves:
🫡 Should be good to ship

PR #246, opened 2025-07-12
Feat: Added Rich Click Themes

Louden7:
Added a simple framework for adding new rich click themes within the rich click package as well as functionality to select the theme from the `@rich_config` decorator within the `RichHelpConfiguration` class.

Features:
- Easily add new themes by setting desired values in the `THEMES` nested dictionary.
- Theme selector is Enum for easy setup
- Theme framework still allows for detailed style overrides when set in the `@rich_config`

Disclamers:
- During pre-commit runs I received a SSL error trying to run `actionlint` (manually ran all others with passes for files modified in the PR)
- Took a stab at some predefined themes (I am not a designer). The themes themselves do not matter much as providing the framework for other users and contributors to easily create and add their own theme to be added into the package.
- Did not modify CHANGELOG or and docs but am happy to add anything necessary.

### DIFF
--- a/examples/12_theme_simple.py
+++ b/examples/12_theme_simple.py
@@ -1,0 +1,72 @@
+import rich_click as click
+from rich_click import rich_config, RichHelpConfiguration
+from rich_click.rich_click_theme import RichClickTheme
+
+@rich_config(RichHelpConfiguration(theme=RichClickTheme.material))
+@click.group()
+@click.option(
+    "--debug/--no-debug",
+    "-d/-n",
+    default=False,
+    help="""Enable debug mode.
+    Newlines are removed by default.
+
+    Double newlines are preserved.""",
+)
+def cli(debug):
+    """
+    My amazing tool does all the things.
+
+    This is a minimal example based on documentation
+    from the 'click' package.
+
+    You can try using --help at the top level and also for
+    specific subcommands.
+    """
+    print(f"Debug mode is {'on' if debug else 'off'}")
+
+
+@cli.command()
+@click.option(
+    "--type",
+    required=True,
+    default="files",
+    show_default=True,
+    help="Type of file to sync",
+)
+@click.option("--all", is_flag=True)
+def sync(type, all):
+    """Synchronise all your files between two places.
+    Example command that doesn't do much except print to the terminal."""
+    print("Syncing")
+
+
+@cli.command(short_help="Optionally use short-help for the group help text")
+@click.option("--all", is_flag=True, help="Get everything")
+def download(all):
+    """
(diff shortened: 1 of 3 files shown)


dwreeves:
Sorry for getting to this a bit late. I'm spending the next month on getting **rich-click** 1.9 out, so I'll be around!

First and foremost: Thank you for your contribution!!!!! ❤️

On first pass, this looks pretty good.

Some notes:

- Users cannot override default of theme with this implementation. E.g. `"style_option": "bold cyan"` would be overridden even if explicitly in kwargs. This is admittedly tedious to solve for in a dataclass, but it would be very nice to have. Probably the easiest solution is to have a sentinel object like `_notset = object()`, and loop over these and set based on theme (with fallback to `default`).

- The xterm colors can be risky because terminal backgrounds can vary quite a bit. I'd be worried about going down that path without first implementing background color checks and having toggles for dark and light backgrounds. For example, on default Ghostty background, the purple looks hard to read:
   image

   Here with Ghostty, I set the bg color to `#ced5eb`, but `term-background`'s approach (Python package) just reads some env vars, then fails to see anything, then just assumes dark because it starts with `xterm`:
   image

   There is a rust lib `termbg` which has a good approach, but this would either need to be given Python bindings or be translated to Python code.
   image image

- Lastly, I want to think about whether this is the right approach going forward. I do think users like setting fast and quick themes like `dracula`, `material`, etc., but I also wonder if there is a way to decouple the colors and the boxes, and for themes that don't use the 16 ANSI system colors, toggle by dark/light.
   - E.g. maybe `dracula` just changes colors and `dracula-slim` changes the colors to dracula and removes boxes (i.e. setting boxes to `SIMPLE`); so every theme would have a `slim` version that condenses the boxes.
   - I'd like, at some point, to implement a theme "builder." Something like: `RichTheme(primary="blue", secondary="green", style="slim")` would be a desirable interface. But also `RichTheme(palette="dracula", style="slim")` as an alternative. **I like the approach of just using strings as well,** because users like things that are simple like strings. So the question is how this all relates.
   - Maybe the solution is, by default strings-as-themes are parsed as: `{palette}` if no hyphen, and `{palette}-{style}` if there is a hyphen? This allows for a consistent feel to strings-as-themes that also lets us implement a more fleshed-out theming system a bit later that allows for customization via setting things like primary/secondary colors.

dwreeves:
Thanks again for your work here!

I'm going to get this in because I think `theme` will be the big selling point of 1.9 (changes to internals aren't very exciting), but note I will be making substantial changes to the implementation here before the 1.9 release.

PR #248, opened 2025-08-01
[WIP] Rich Parameters + Panels

dwreeves:
# High-Level Objectives

- Silently deprecate rich-click "groups"; rename to panel (maintain 100% backwards compatibility). This is a better namespace.
- Introduce new shimmed decorators for parameters that implement `RichParameter`s, with two key distinctions: (1) `help=` for arguments (2) 
- Introduce `click.panel()` decorator as new API for panels.
- Test coverage needs to go up as part of implementing all of this, e.g. help option, version option, etc. need full coverage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 
 Big changes:
 
-- TODO
+- `help=` is now a valid kwarg for `@argument()` decorator. See docs for more information.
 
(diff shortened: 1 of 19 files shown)


dwreeves:
Tests (mostly) done, just need to set default behavior for `click.argument(help=)` and good to merge.

PR #252, opened 2025-08-10
1.9 - Panels

dwreeves:
Not done, but very close.


dwreeves:
Shaping up to be in a good spot but need to save the rest for tomorrow and the coming days.

Minor behaviors to fix: Default commands panel is above custom panels but should be below it.

Also, many tests are required, asserting order of things in stress-tested situations.

PR #254, opened 2025-08-14
[draft] rich-click CLI fixes

dwreeves:
Something fishy about how the rich-click CLI works in all honesty.

Since `format_options()` calls `format_commands()`, it doesn't seem that `format_commands()` should need to be directly called. I imagine what is happening is that `RichCommand.format_options` is being called rather than `RichGroup.format_options`. This needs to be fixed.

(Also, there was a minor bug in how the epilog was placed before `format_commands`. That is fixed here as well.)

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -13,5 +13,5 @@ jobs:
           python-version: 3
       - name: Install dependencies
         run: |
-          pip install --editable ".[dev]"
+          pip install ".[dev]"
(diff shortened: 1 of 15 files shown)


dwreeves:
Also an integration test failure with `1.9.0dev0`:

```python
uv run --with litestar --with "rich-click==1.9.0dev0" rich-click litestar --help
```

Subcommands don't render:

image

dwreeves:
Everything is fixed and perfect.

PR #259, opened 2025-08-16
Themes

dwreeves:


dwreeves:
Tables are now close to rendering the way they used to but still not quite the same. I'm very unclear as to why it changed all of a sudden, especially given that now I am using `Text()` objects when `Columns()` aren't necessary.

New behavior:


I'm not going to update unit-tests until I have an explanation for the change that I am satisfied with. Not really a big deal if I need to update most of these tests. The only one I saw with a major dramatic change was `test_metavars.py`

dwreeves:
Closing this PR as it is outdated and will be redoing it.

PR #270, opened 2025-08-22
first implementation

wr1:
First concept for tree view, bit rough still. 

-  example 13 works as intended
-  example 14 doesn't, requires explicit setting of Command and Group class to work


dwreeves:
Thank you so much for this speedy PR and I'm super excited to get it in for **1.9.0**.

It looks like the main test that is failing here is a test that asserts **`rich` is never imported during normal CLI execution.** Basically, when a user wants to run their program, when they are _not_ rendering help text, `rich` should never be inside of `sys.modules`.

The reason we assert this is because importing rich slows down code execution by a few milliseconds, and increases the memory footprint. And we pride ourselves on doing the best we can to make sure we get in the way of code execution as little as possible 😄 See this:

image

image

---

The way we get around this is through lazy imports, and you'll see examples of this throughout the code. A few examples: 

Here is an example of a lazy import direct to rich:

https://github.com/ewels/rich-click/blob/b77ce9ebc11a59e9a547221ea0d2df74f82a8563/src/rich_click/rich_help_formatter.py#L40-L52

This example here is a lazy import to the `rich_help_rendering` module. The `rich_help_rendering` module _eagerly_ loads rich but the module itself is never eagerly loaded.

https://github.com/ewels/rich-click/blob/b77ce9ebc11a59e9a547221ea0d2df74f82a8563/src/rich_click/rich_command.py#L265-L268

dwreeves:
Hi,

Some slightly bad news. I don't think we can get this in for 1.9. There's just too much work to do cleaning up 1.9 and getting this ready will take up too much additional time.

But, `--tree` will definitely be one of the bigger features for a 1.10 release!

Thank you a ton for your work on this. Please keep this PR open and we'll work on integrating it at some point in the near to medium term future, and it will get its own release.

PR #282, opened 2025-09-02
[DRAFT] Add Typer support

dwreeves:
image

@ewels Given the history of this library and Typer, I believe we've gone full circle. 😏


dwreeves:
pre-commit checks should not be passing as the typing is a complete mess. It seems to only be checking files changed in latest commit.

PR #284, opened 2025-09-06
Very minor bugfixes

dwreeves:
Sooooooo close!


dwreeves:
I think this is actually it...

PR #289, opened 2025-09-15
1.9.0 full release

dwreeves:
ready? 🫡

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Version 1.9.0.dev6 (WIP)
+## Version 1.9.0 (2025-09-15)
 
 > [!WARNING]
 > Version 1.9.0 deprecates support for a lot of old versions of things:
--- a/docs/blog/posts/version-1.9.md
+++ b/docs/blog/posts/version-1.9.md
@@ -1,5 +1,5 @@
 ---
-draft: true
+draft: false
 slug: version-1.9
 date: 2025-09-15
 authors:
--- a/src/rich_click/__init__.py
+++ b/src/rich_click/__init__.py
@@ -8,7 +8,7 @@ customization required.
 from __future__ import annotations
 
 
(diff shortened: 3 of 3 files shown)


ewels:
One last thing.. I was thinking about your integration tests, it's a really nice idea. I checked ClickPy stats to see who the most-trafficked consumers are of the library: https://clickpy.clickhouse.com/dashboard/rich-click

It looks like ScyPy is the biggest, [since their v1.9.0 release](https://github.com/scipy/scipy/blob/d13bb11a21ffcc3bf9bbbaf687c59d4dd8080e4b/doc/source/release/1.9.0-notes.rst#other-changes). Can see it being pulled in [here](https://github.com/schettino72/scipy/blob/4ca1ec294c7a729c7d8abdb2eb35bde9649ca2d0/environment.yml#L45-L46). Though... wait... it was removed again 😆  https://github.com/scipy/scipy/issues/22887 ok, so not such a big deal. (Maybe could offer to help out with spin - see https://github.com/scientific-python/spin/pull/23)

Others are apache-airflow (I guess you tried this already?) and [beartype](https://github.com/beartype/beartype), then it starts to drop off a bit.

The only other project that I have that uses rich-click extensively is [nf-core/tools](https://github.com/nf-core/tools).

Anyway, just a side thought! Not sure that this needs to be a blocker, I was thinking it was worth testing SciPy but if they're not using rich-click then I think we're probably fine to release.

dwreeves:
@ewels There is nothing more that needs to be done on our end.

There is one bug that was _just_ recently uncovered in `esptool`, a big library that uses **rich-click**.

... Unfortunately, this is technically a bug on their end, not ours, and there just isn't anything we can do about it. I am attempting to assist by providing a PR. https://github.com/espressif/esptool/pull/1121

I think what we should do is **merge this, but don't click the release button just yet.** We should give the `esptool` maintainers a few hours to do a patch release that implements a fix. The maintainer of the repo is online right now and should get to it soon.

PR #326, opened 2026-03-02
Add RichCliRunner for testing Rich prompt/output capture

### LINKED ISSUE #232: Feature request: provide modified `CliRunner`
I'm working to replace our use of Click UI utilities with Rich equivalents. However, I ran into an issue today that prevents me from fully switching over. In particular I cannot yet use [`rich.prompt.Prompt`](https://rich.readthedocs.io/en/stable/reference/prompt.html#rich.prompt.Prompt) & [`rich.prompt.Confirm`](https://rich.readthedocs.io/en/stable/reference/prompt.html#rich.prompt.Confirm) in place of [`click.prompt`](https://click.palletsprojects.com/en/stable/api/#click.prompt) & [`click.confirm`](https://click.palletsprojects.com/en/stable/api/#click.confirm), respectively.

During tests, the `CliRunner` that Click provides [temporarily reassigns](https://github.com/pallets/click/blob/8.1.8/src/click/testing.py#L311-L320) internal utilities in order to make test assertions work properly. Using the utilities provided by Rich aren't aware of what's happening so assertions break both from expected output and from checks within standard library functions like [this line](https://github.com/python/cpython/blob/3.13/Lib/getpass.py#L98) checking if stdin has been switched since the interpreter started.

I'm not sure if this is in scope but I think it should be as I view this library as the foundational project that allows Rich to provide all UI functionality while using Click only for CLI definitions/parsing 😄

mirs002:
## Summary
- add `rich_click.testing.RichCliRunner`, a `click.testing.CliRunner` subclass
- rebind Rich's global cached console during `CliRunner.isolation()` so output goes to Click's isolated stdout/stderr streams
- restore the previous global Rich console after invocation to avoid leaking test state
- export `RichCliRunner` from `rich_click`
- add regression tests covering prompt capture behavior with a preconfigured global Rich console

## Why
When Rich's global console has already been initialized before `CliRunner.invoke()`, prompt / console output may go to the pre-existing console stream instead of Click's test-captured streams. This makes asserting Rich prompt output difficult in tests.

This change provides a dedicated runner for Rich-based CLIs that keeps Rich aligned with Click's test isolation.

Closes #232.

### DIFF
--- a/src/rich_click/__init__.py
+++ b/src/rich_click/__init__.py
@@ -91,6 +91,7 @@ from rich_click.rich_panel import RichPanel as RichPanel
 from rich_click.rich_parameter import RichArgument as RichArgument
 from rich_click.rich_parameter import RichOption as RichOption
 from rich_click.rich_parameter import RichParameter as RichParameter
+from rich_click.testing import RichCliRunner as RichCliRunner
 
 from . import rich_click as rich_click
 
--- a/src/rich_click/patch.py
+++ b/src/rich_click/patch.py
@@ -26,7 +26,6 @@ __TyperOption: Type["typer.core.TyperOption"]
 
 
 class _PatchedTyperContext(RichContext):
-
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         RichContext.__init__(self, *args, **kwargs)
 
--- a/src/rich_click/rich_command.pyi
+++ b/src/rich_click/rich_command.pyi
@@ -41,7 +41,6 @@ G = TypeVar("G", bound=click.Group)
 OVERRIDES_GUARD: bool = False
 
 class RichCommand(click.Command):
-
     context_class: Type[RichContext] = RichContext
     _formatter: Optional[RichHelpFormatter] = None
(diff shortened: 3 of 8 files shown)


mirs002:
## Note on the failing `test (3.10, 8.1.*, 14.*)` job

Just a heads-up that the one remaining CI failure is unrelated to this PR.

The job fails with an `ImportError` in `conftest.py` when importing Typer — before any of the new code in this PR is even loaded:

```
from typer.testing import CliRunner as TyperCliRunner
...
TypeError: 'type' object is not subscriptable
```

This is caused by **Typer 0.24.x** (released recently) dropping support for the `Python 3.10 + Click 8.1` combination. The same job was passing on the last push to `main` on Jan 31 — before Typer 0.24 was released.

You can verify this by re-running the `main` branch CI — it will reproduce the same failure independently of this PR.

Happy to update the PR once this is resolved upstream or the matrix is adjusted! 🙂

dwreeves:
Thanks! I'll review when I get a moment.

PR #329, opened 2026-05-28
Fix regression in patching and version bump

### LINKED ISSUE #328: Patching broken with Click 8.4.0
```
TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```

Root cause appears to be that `Parameter` is now an `ABC` subclass, and `ABCMeta` conflicts with `PatchMeta`.

dwreeves:
Resolves #328


dwreeves:
Coming back to our Click + Typer patching after 6 months feels like this.

image

PR #331, opened 2026-06-02
Machine-readable `--help-json`

ewels:
CLIs built on rich-click are increasingly consumed not just by humans but by tooling and LLM agents. Those consumers struggle with the rendered `--help` screen: it's laid out for human reading, wraps to terminal width, and carries Rich styling that obscures the underlying structure.

rich-click already has all of this information as structured data. This PR exposes it directly.

This PR adds a single opt-in config option, `help_json`. When enabled, **every** command and group gains a global `--help-json` flag:

```python
import rich_click as click
click.rich_click.HELP_JSON = True
```

`cli --help-json` prints the current command's help, usage and full parameter detail as JSON, plus a recursive index of subcommand names:

```json
{
  "name": "cli",
  "path": "cli",
  "help": "My CLI.",
  "usage": "cli [OPTIONS] COMMAND [ARGS]...",
  "params": [],
  "subcommands": { "hello": {} }
}
```

Design notes:

- **Contextual / progressive disclosure.** The flag dumps the current command plus a names-only index of its descendants, so agents drill down one level at a time instead of pulling the whole tree into context at once.
- **Eager**, like `--help`, so `cli <cmd> --help-json` works even when required arguments are missing.
- **Name is configurable** via `help_json_option_name` (defaults to `--help-json` because many CLIs already define their own `--json` data-output flag).
- Off by default (for now), so zero behavioural change unless you opt in.

This PR ports [nf-core/tools#4310](https://github.com/nf-core/tools/pull/4310) down to the rich-click level, so any package built on rich-click can opt in with one config option rather than reimplementing the introspection itself.


dwreeves:
@ewels I'll take a closer look this weekend if that's alright.

One thought / question that comes to mind before a more thorough review-- Click 8.x added a `to_info_dict()` method to CLIs which prints out the structure of a CLI as a dict, which in turn can be serialized to json. Do you think this would be more appropriate or do you have a preference for your own schema?

ewels:
Absolutely - no rush!

This PR uses that Click function, I should probably look a bit more carefully at what that output actually looks like in fairness.

I was aiming for JSON-schema output because I figured it was a common format and LLMs would understand how to interpret it easily. But I don't feel particularly strongly about it.

dwreeves:
Got it. I see it is using to_info_dict upon closer review for param to dict (sorry, skimming on my phone... need to sit down and look) but what I saw through the skim is that it is doing some minor schema renaming: `command_schema()` is all custom, and the param to dict function calls to_info_dict but also renames some of the fields.

Basically, if someone said "--help-json in Click" I would have assumed the implementation would be a straight `print(json.dumps(cmd.to_info_dict()))`.

It may be helpful to understand if Click had any deep justification for why they did things the way they did things with to_info_dict(). Perhaps they do, perhaps not. That said an advantage of hooking directly into to_info_dict() is to allow for better user control with method overrides. (I mean, if anyone cares, but why not make the code beautiful and composable?) Lastly, even if we do renames, a question arises of: do we detect "extra" fields in to_info_dict() and populate those? E.g. user adds `foo: "bar"` to to_info_dict(); should that occur in the `--help-json`?

I think the feature makes sense overall, great addition. Again, I'll take a closer look this weekend.

Thanks for coming back! Heh.

ewels:
Hello! So I had a sit down and a play with this PR again with a bit more thought.

> I would have assumed the implementation would be a straight `print(json.dumps(cmd.to_info_dict()))`

The main problem with this is that it would print the _entire_ CLI help in one go. I wanted to keep this output as lean as possible to be token-efficient, with progressive disclosure. That's why the output gives a full description of the current command but for the nested subcommands (recursively) only lists their name + a one-line description, not their full params/usage (I didn't describe this well in the PR description sorry).

Also, the click output prints a tonne of field names, most of which will be empty most of the time (`"secondary_opts"=[]`, `"prompt": null` etc etc). Again wastes tokens and makes the output less clear for basically no gain.

All this adds up, and so a typical top-level `--help-json` for the two approaches ends up looking very different:

<details><summary>Custom output (26 lines, 576 chars)</summary>

```json
{
  "name": "cli",
  "path": "cmp.py",
  "help": "A demo CLI.",
  "usage": "cmp.py [OPTIONS] COMMAND [ARGS]...",
  "params": [
    {
      "name": "verbose",
      "kind": "option",
      "opts": ["--verbose", "-v"],
      "type": "Bool",
      "is_flag": true,
      "help": "Enable verbose output."
    }
  ],
  "subcommands": {
    "db": {
      "help": "Manage the database.",
      "subcommands": {
        "migrate": { "help": "Run migrations." },
        "rollback": { "help": "Roll back to a revision." }
      }
    },
    "hello": { "help": "Greet someone." }
  }
}
```

</details>

<details><summary>Click output (254 lines, 7,009 chars)</summary>

```json
{
  "name": "cli",
  "params": [
    {
      "name": "verbose",
      "param_type_name": "option",
      "opts": ["--verbose", "-v"],
      "secondary_opts": [],
      "type": { "param_type": "Bool", "name": "boolean" },
      "required": false,
      "nargs": 1,
      "multiple": false,
      "default": false,
      "envvar": null,
      "help": "Enable verbose output.",
      "prompt": null,
      "is_flag": true,
      "flag_value": true,
      "count": false,
      "hidden": false
    },
    {
      "name": "help",
      "param_type_name": "option",
      "opts": ["--help"],
      "secondary_opts": [],
      "type": { "param_type": "Bool", "name": "boolean" },
      "required": false,
      "nargs": 1,
      "multiple": false,
      "default": false,
      "envvar": null,
      "help": "Show this message and exit.",
      "prompt": null,
      "is_flag": true,
      "flag_value": true,
      "count": false,
      "hidden": false
    }
  ],
  "help": "A demo CLI.",
  "epilog": null,
  "short_help": null,
  "hidden": false,
  "deprecated": false,
  "commands": {
    "db": {
      "name": "db",
      "params": [
        {
          "name": "help",
          "param_type_name": "option",
          "opts": ["--help"],
          "secondary_opts": [],
          "type": { "param_type": "Bool", "name": "boolean" },
          "required": false,
          "nargs": 1,
          "multiple": false,
          "default": false,
          "envvar": null,
          "help": "Show this message and exit.",
          "prompt": null,
          "is_flag": true,
          "flag_value": true,
          "count": false,
          "hidden": false
        }
      ],
      "help": "Manage the database.",
      "epilog": null,
      "short_help": null,
      "hidden": false,
      "deprecated": false,
      "commands": {
        "migrate": {
          "name": "migrate",
          "params": [
            {
              "name": "steps",
              "param_type_name": "option",
              "opts": ["--steps"],
              "secondary_opts": [],
              "type": { "param_type": "Int", "name": "integer" },
              "required": false,
              "nargs": 1,
              "multiple": false,
              "default": 1,
              "envvar": null,
              "help": "How many migrations to apply.",
              "prompt": null,
              "is_flag": false,
              "flag_value": null,
              "count": false,
              "hidden": false
            },
            {
              "name": "help",
              "param_type_name": "option",
              "opts": ["--help"],
              "secondary_opts": [],
              "type": { "param_type": "Bool", "name": "boolean" },
              "required": false,
              "nargs": 1,
              "multiple": false,
              "default": false,
              "envvar": null,
              "help": "Show this message and exit.",
              "prompt": null,
              "is_flag": true,
              "flag_value": true,
              "count": false,
              "hidden": false
            }
          ],
          "help": "Run migrations.",
          "epilog": null,
          "short_help": null,
          "hidden": false,
          "deprecated": false,
          "panels": [],
          "aliases": []
        },
        "rollback": {
          "name": "rollback",
          "params": [
            {
              "name": "revision",
              "param_type_name": "argument",
              "opts": ["revision"],
              "secondary_opts": [],
              "type": { "param_type": "String", "name": "text" },
              "required": true,
              "nargs": 1,
              "multiple": false,
              "default": null,
              "envvar": null,
              "help": null
            },
            {
              "name": "help",
              "param_type_name": "option",
              "opts": ["--help"],
              "secondary_opts": [],
              "type": { "param_type": "Bool", "name": "boolean" },
              "required": false,
              "nargs": 1,
              "multiple": false,
              "default": false,
              "envvar": null,
              "help": "Show this message and exit.",
              "prompt": null,
              "is_flag": true,
              "flag_value": true,
              "count": false,
              "hidden": false
            }
          ],
          "help": "Roll back to a revision.",
          "epilog": null,
          "short_help": null,
          "hidden": false,
          "deprecated": false,
          "panels": [],
          "aliases": []
        }
      },
      "chain": false,
      "panels": [],
      "aliases": []
    },
    "hello": {
      "name": "hello",
      "params": [
        {
          "name": "count",
          "param_type_name": "option",
          "opts": ["--count"],
          "secondary_opts": [],
          "type": { "param_type": "Int", "name": "integer" },
          "required": false,
          "nargs": 1,
          "multiple": false,
          "default": 1,
          "envvar": null,
          "help": "Number of greetings.",
          "prompt": null,
          "is_flag": false,
          "flag_value": null,
          "count": false,
          "hidden": false
        },
        {
          "name": "fmt",
          "param_type_name": "option",
          "opts": ["--fmt"],
          "secondary_opts": [],
          "type": { "param_type": "Choice", "name": "choice", "choices": ["json", "yaml"], "case_sensitive": true },
          "required": false,
          "nargs": 1,
          "multiple": false,
          "default": "json",
          "envvar": null,
          "help": "Output format.",
          "prompt": null,
          "is_flag": false,
          "flag_value": null,
          "count": false,
          "hidden": false
        },
        {
          "name": "name",
          "param_type_name": "argument",
          "opts": ["name"],
          "secondary_opts": [],
          "type": { "param_type": "String", "name": "text" },
          "required": true,
          "nargs": 1,
          "multiple": false,
          "default": null,
          "envvar": null,
          "help": null
        },
        {
          "name": "help",
          "param_type_name": "option",
          "opts": ["--help"],
          "secondary_opts": [],
          "type": { "param_type": "Bool", "name": "boolean" },
          "required": false,
          "nargs": 1,
          "multiple": false,
          "default": false,
          "envvar": null,
          "help": "Show this message and exit.",
          "prompt": null,
          "is_flag": true,
          "flag_value": true,
          "count": false,
          "hidden": false
        }
      ],
      "help": "Greet someone.",
      "epilog": null,
      "short_help": null,
      "hidden": false,
      "deprecated": false,
      "panels": [],
      "aliases": []
    }
  },
  "chain": false,
  "panels": [],
  "aliases": []
}
```

</details>

This means I have to dig in and start messing with the output, and once you go from nothing to something, the rest kind of follows quite quickly.

> Schema used

I went back and forth on trying to use JSON schema (in the above comment I said it was using it already - that wasn't actually true, I'd deviated and forgotten). In the commit history there is a version that does proper JSON schema with a smattering of custom keys but it was a little more verbose / unclear and I wasn't convinced that it really enabled any additional functionality. LLMs should be able to do fine with the custom schema.

So yeah, as it stands, it is a custom schema that is optimised for readability and succinctness.

> Ability to customise _(why not make the code beautiful and composable?)_

Yep, agreed - no reason not to have all the nice things. In the latest commits I've reworked it so that the output picks up any extra keys automatically. So your `foo: "bar"` shows up:

```python
class MyCommand(click.RichCommand):
    def to_info_dict(self, ctx):
        info = super().to_info_dict(ctx)
        info["examples"] = ["mytool deploy prod"]  # -> top-level "examples"
        return info
```

```json
{
  "name": "deploy",
  "path": "mytool deploy",
  "help": "Deploy the app.",
  "usage": "mytool deploy [OPTIONS] TARGET",
  "params": [
    { "name": "target", "kind": "argument", "type": "String", "required": true }
  ],
  "examples": ["mytool deploy prod"]
}
```

Custom command keys are printed at the top level, custom param keys on the param. There's also a `help_json_transform` hook for people, to avoid subclassing.

> It may be helpful to understand if Click had any deep justification for why they did things the way they did things with `to_info_dict()`

I had a poke around, seems like it dates back to 2020 and was mostly intended for documentation websites. It seems to basically just be a dump rather than anything massively intentional.

dwreeves:
Sorry, been a little busy 😭 I haven't forgotten to give this a real review!

ewels:
No rush!

src/rich_click/rich_command.py
```diff
@@ -76,6 +76,8 @@ def __init__(
         self.aliases: Iterable[str] = aliases or []
         if not hasattr(self, "_help_option"):
             self._help_option = None
+        if not hasattr(self, "_help_json_option"):
+            self._help_json_option: Optional[click.Option] = None
```
dwreeves: I don't think this should be necessary.

I should have commented it as such, but `if not hasattr(self, "_help_option"):` line is a Click compatibility thing. I believe `self._help_option` was added in Click 8.1.8. So that hacky if statement is just to support previous versions of Click.

Here we can just be a little more plain with it:

```suggestion
        self._help_json_option: Optional[click.Option] = None
```

src/rich_click/rich_click.py
```diff
@@ -171,6 +172,11 @@
 OPTION_GROUPS: Dict[str, List[OptionGroupDict]] = {}
 USE_CLICK_SHORT_HELP: bool = False  # Use click's default function to truncate help text
 HELPTEXT_SHOW_ALIASES: bool = True
+# Add a global `--help-json` flag to every command, printing machine-readable help as JSON (for LLMs/tooling)
+HELP_JSON: bool = False
+HELP_JSON_OPTION_NAME: str = "--help-json"  # Flag name used when HELP_JSON is enabled
```
dwreeves, line 178: This is fine. Another option would be to treat this similarly to how help options work in click, i.e.

```python
@click.command(
    context_settings=dict(
        help_option_names=["--help", "-h"],
        help_json_option_names=["--help-json"]
    )
)
def cli():
    """My CLI"""
```

This has more parallelism, also it has some advantage of pushing the config to be more style than functional. (Right now there is some bleeding between the two concepts of the config being for style only vs functional.)

I think the main downside is that it is somewhat expected that the "Rich Click" stuff is inside of the "Rich config." So by having inside of the context settings, we are sprawling the places where we add stuff and sprawling what users would be expected to touch.

Not proposing any changes! Just making sure a different potential design is considered.

src/rich_click/help_json.py
```diff
@@ -0,0 +1,244 @@
...
+
+
+if TYPE_CHECKING:  # pragma: no cover
+    from rich_click.rich_parameter import RichOption
+
+
+DEFAULT_HELP_JSON_OPTION_NAME = "--help-json"
+DEFAULT_HELP_JSON_HELP = "Print this command's help/usage as JSON."
+
+#: Type of the optional ``help_json_transform`` hook: ``(schema, command, ctx) -> schema``.
```
dwreeves, line 44: I feel like this would make more sense as something which could be handled via subclassing the `RichCommand` and overriding a method rather than as a random callable.

This actually leads me to my biggest API suggestion for this code.

What Click does is when calling `--help`, it calls `Command.format_help(ctx, formatter) -> None`.

I believe that `--help-json` should behave roughly in a similar way. Maybe something like this:

```python
class RichCommand:

    def get_help_json(self, ctx: RichContext) -> str:
        formatter = ctx.make_formatter()
        data = self.format_help_json(ctx, formatter)
        return json.dumps(data)

    def format_help_json(self, ctx: RichContext, formatter: RichHelpFormatter) -> dict:
        # Logic goes here to construct dict.
        ...
```

This more or less mirrors base Click's `get_help` and `format_help` pattern, except that in Click, the formatter stores data in the buffer whereas here I'm proposing just returning the value statelessly (the formatter in this case would just be purely for the config; also possibly to handle some other niche situations).

dwreeves (commented):
Alright, mostly looks good. Thanks for your patience as I find time to review this!

I have one really big API design comment-- I personally think it would make sense to mirror how Click does things more with normal help text (not only is it a pretty reasonable design, but also there is a "least astonishment" reason for doing so). See the comments below for more info on my proposal for two methods to add to `RichCommand`.

One thing not mentioned in my comments: **How to handle the rich-click CLI.** Because right now, there are no changes to the rich-click CLI, but I believe we could make a few changes. A couple ideas come to mind:

- We can add a `rich-click --output json [command]`. For example, `rich-click --output json flask --help` returns json output. This should be more or less equivalent to `rich-click -c '{"help_json": True}' flask --help-json`.
- Another question is what to do if we do something like: `rich-click -c '{"help_json": True}' --output svg flask --help-json`. Should this render an SVG output of the help JSON? I believe that it should!

src/rich_click/rich_click.py
ewels: Yeah I like your suggestion. Maybe we can have both? Mine as a fall-back, as it'd be for all commands.

ewels:
Pushed 0b2e223 addressing the review feedback. Summary:

Mirror click's `get_help`/`format_help` - added two overridable methods on `RichCommand`:
- `get_help_json(ctx) -> str` — makes a formatter, calls `format_help_json`, serializes.
- `format_help_json(ctx, formatter) -> dict` — builds the schema statelessly and applies the `help_json_transform` hook.

The `--help-json` callback now just calls `ctx.command.get_help_json(ctx)`, so the schema/transform/serialize logic lives on the command and is overridable by subclassing. `command_schema` remains the underlying builder. So a subclass can do:

```python
class MyCommand(click.RichCommand):
    def format_help_json(self, ctx, formatter):
        data = super().format_help_json(ctx, formatter)
        data["version"] = "1.2.3"
        return data
```

I added `help_json_option_names` to `context_settings`, parallel to click's `help_option_names`. Setting it both enables and names the flag and takes precedence over the config; the `help_json` / `help_json_option_name` rich config stays as the all-commands fallback.

```python
@click.command(context_settings={"help_json_option_names": ["--schema"]})
def cli():
    ...
```

rich-click CLI `--output json`: I put this in a stacked PR so as to keep the two easier to review. See #335:

- #335

src/rich_click/rich_click.py
dwreeves: ~~Having this be in multiple places would violate the zen of Python `There should be one-- and preferably only one --obvious way to do it.` I would suggest having one way and sticking to it, whether it's rich config or context settings. Maybe just stick with what we have?~~ Eh. Doesn't matter 😆

dwreeves:
Good change, thanks! I'll review more thoroughly again this weekend but I imagine this is pretty close to what it should be if not done. Great work!

src/rich_click/rich_click.py
ewels: My main motivation for both was that if you have many commands then setting it once is easier / less code. If you only have one and no rich-click config then the decorator is nicer. So I could see both scenarios being preferred by different people.

But yeah honestly I don't mind and am happy to take the pure approach if you would rather :)

dwreeves: Hah I really don't care. 😂 My motivation was just to make sure all the alternatives for the API design were considered. After all, internals can change, high level APIs are forever. I don't have any preference!

ewels:
Added a footer tip to `--help` that advertises `--help-json` when it's enabled, so both users and agents discover the flag without having to scan the options list:

```
Tip: add --help-json to any command for machine-readable help.
```

Rendered in dim text and shown only when `--help-json` is active (it reflects a custom flag name too). Configurable via `help_json_show_tip` (toggle), `help_json_tip_text` (wording, `{}` → flag name) and `style_help_json_tip` (style, default `dim`). Docs and tests included.

ewels:
I found that `_param_to_dict()` listed a number of keys in `_STANDARD_PARAM_KEYS` but never actually mapped them into the output.

- `secondary_opts`: negation flags like `--no-debug` for a `--debug/--no-debug` option (previously only `--debug` appeared)
- `envvar`: the env var(s) backing an option
- `nargs`: surfaced when `!= 1`, so variadic args (`-1`) and fixed multi-value options (`N`) are distinguishable from single-value params
- `count`: `-vvv`-style counters, distinct from a plain boolean flag
- `flag_value`: for value-flags that share a destination (`--upper`/`--lower`); suppressed for plain bool flags where it's just a noisy `True`
- `prompt`: the interactive prompt string

I also added a nested `type_info` object carrying the remaining type constraints that were being compacted away: `IntRange`/`FloatRange` `min`/`max`, `DateTime` `formats`, `Path` flags (`exists`, `dir_okay`, …), `Choice` `case_sensitive`.

ewels:
Ok I started thinking more about this feature and have a new branch where I built it out into something a bit different / a lot more extensive (and I think, more useful).

I'll do that as a separate PR in a moment, moving this PR to draft now as I don't want to merge anything until we have a feel for what direction we want to go in with this.

_[Edit: See #337]_

ewels:
Closing in favour of #337

PR #332, opened 2026-06-02
Modernize codebase to Python 3.10+

ewels:
Slightly more controversial PR.. Opening at least partially just for discussion.

Drops support for Python 3.8 and 3.9 (now both [EOL](https://devguide.python.org/versions/)); minimum is now Python 3.10. This allows us to clean up the code quite a bit:

- **Typing syntax:** PEP 604 unions (`X | None` instead of `Optional`/`Union`) and PEP 585 builtin generics (`list`, `dict`, `tuple`, `type`) throughout, with ABCs imported from `collections.abc`.
- **Removed shims** that only existed for Python &lt;3.10:
  - the `entry_points()` 3.7 compat wrapper in `cli.py`
  - the `Concatenate`/`ParamSpec` `typing_extensions` fork in `decorators.py`
  - the `&lt;3.9` skip in `tests/test_types.py`
- **Tooling:** added the `UP` (pyupgrade) ruff rule and set `ruff`/`pyright` targets to py310.
- **CI:** dropped 3.8/3.9 from the test matrices (`pytest`, `scheduled-tests`, `test-examples`).
- **Misc:** `subprocess.run(stdout=PIPE, stderr=PIPE)` → `capture_output=True` in test helpers.

Merged `main` in, which resolved against the dev-tooling overhaul that landed since this branch was opened:

- The pre-commit → prek switch: `.pre-commit-config.yaml` and the old workflow are gone, and `prek.toml` already pins a newer ruff than this branch bumped to.
- Dev/docs requirements moved to dependency groups, so the `optional-dependencies` edits here are obsolete; the dev group's `mypy` marker is simplified now that the floor is 3.10.
- `[tool.black]` dropped (main replaced Black with `ruff format`), and the `warn_unused_ignores = false` workaround is no longer needed — the dev group pins `mypy>=2.3`, so a single set of `# type: ignore`s no longer has to span several mypy versions.
- Kept main's versions of the `RichContext.__enter__`/`__exit__` removal and the panel table filtering fix.
- `RichHelpFormatter.write_dl` takes `Iterable` again, to stay compatible with the click supertype.

`prek run -a` and the test suite pass, on the dev env (3.11) and on 3.10.


dwreeves:
Not too controversial, end of life is end of life and rich-click has been overly generous about supporting old Python versions. I'll look over this weekend but I imagine it's pretty straightforward.

dwreeves (approved):
Seems fine to me.

The one thing I would say is that removing support for a Python version makes more sense to me as a minor version release, not a patch release, so this should correspond with that.

ewels:
Thanks for checking this over @dwreeves 

Let's merge this in. Feels like a nice complement to the dev tooling updates, and good to get it all done in one pass (more or less).

PR #334, opened 2026-06-16
Fix RichContext errors_in_output_format inheritance guard

ewels:
## What

`RichContext.__init__` inherits a handful of attributes from the parent context. The guard for `errors_in_output_format` checked the wrong attribute name:

```python
if errors_in_output_format is None and hasattr(parent, "export_console_as"):  # wrong attr
    self.errors_in_output_format = parent.errors_in_output_format
```

It should check `hasattr(parent, "errors_in_output_format")`, matching the sibling blocks for `help_to_stderr` and `export_console_as`.

## Impact

Behaviourally neutral today: both `export_console_as` and `errors_in_output_format` are class-level defaults, so the `hasattr` check on a `RichContext` parent is always `True` either way and inheritance still happens. The guard was only correct by accident — this corrects the intent and makes it robust if those attributes ever become instance-only.

## Test

Adds `test_child_context_inherits_errors_in_output_format`, locking in the inheritance contract.

Spotted while reviewing #331.

### DIFF
--- a/src/rich_click/rich_context.py
+++ b/src/rich_click/rich_context.py
@@ -61,7 +61,7 @@ class RichContext(click.Context):
         else:
             self.export_console_as = export_console_as or self.export_console_as
 
-        if errors_in_output_format is None and hasattr(parent, "export_console_as"):
+        if errors_in_output_format is None and hasattr(parent, "errors_in_output_format"):
             self.errors_in_output_format = parent.errors_in_output_format  # type: ignore[union-attr]
         else:
             self.errors_in_output_format = errors_in_output_format or self.errors_in_output_format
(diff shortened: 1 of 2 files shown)


dwreeves (approved):
Ah good catch. LGTM.

The reason we do these `hasattr()` things is to support interplay between Click and rich-click subcommands, e.g.

```python
import rich_click
import click

@click.group()
def foo(): ...

@rich_click.command()
def bar(): ...

foo.add_command(bar)
```

Here the parent Context will be of type`click.Context` but not `RichContext`, and thus a duck-typed attr check will fail.

What ended up happening I believe was that I renamed something, but didn't catch the `hasattr()` because LSPs don't recognize these sort of string-y situations. Woops!

ewels:
All good! I can't take any credit for this really, it was all our AI overlords 😅

PR #335, opened 2026-06-16
rich-click CLI: --output json for machine-readable help

ewels:
Stacked on top of #331 (base branch `help-json`). Implements dwreeves's CLI suggestion from that review.

## What

Extends the rich-click CLI's `--output` option with a `json` value, so machine-readable help works through the CLI wrapper for *any* Click CLI — even ones that haven't opted into `help_json`.

```console
$ rich-click --output=json flask --help
{ "name": "flask", "path": "flask", ... }
```

Two behaviours:

1. **`rich-click --output=json [cmd] --help`** emits the `--help-json` schema instead of rendered help. A plain `--help` yields JSON — no need for the target to enable `help_json`. Implemented via a `RichCommand.get_help()` override that returns `get_help_json()` when the context's `export_console_as` is `"json"`.
2. **`--output` applies to the `--help-json` flag too.** The schema is rendered through the recording console and exported, so `rich-click -c '{"help_json": true}' --output=svg [cmd] --help-json` produces an SVG of the JSON (dwreeves's second case).

## Why `--output json`, not `--json`

It's a *value* of the existing `-o`/`--output` option, not a new flag. The wrapper consumes `--output` before the script args (`allow_interspersed_args=False`), so the wrapped command never sees it — no risk of clashing with the target's own options.

## Notes

- `getvalue()` treats `"json"` as plain-text export defensively; it should never reach there for help, since `get_help` short-circuits.
- Docs updated (`rich_click_cli.md`, cross-ref from `machine_readable_help.md`); 3 new tests + regenerated `--help` snapshot.

⚠️ Review/merge #331 first — this is stacked on it, so the diff will look correct once the base lands on `main`.

### DIFF
--- a/docs/documentation/machine_readable_help.md
+++ b/docs/documentation/machine_readable_help.md
@@ -18,6 +18,8 @@ This adds a global `--help-json` flag to every command and group:
 
 See the [Configuration](configuration.md) page for how to set config options globally or per-command with the `rich_config` decorator.
 
+You can also get JSON help for _any_ Click CLI — even one that hasn't opted in — with the [rich-click CLI](rich_click_cli.md#render-help-as-json): `rich-click --output=json [command] --help`.
+
 ## Example output
 
 Running the top-level command with `--help-json` prints the current command's help, usage and full parameter detail as JSON, together with a recursive index of subcommand names:
--- a/docs/documentation/rich_click_cli.md
+++ b/docs/documentation/rich_click_cli.md
@@ -44,6 +44,27 @@ You can also use `rich-click --output=html [command]` to render rich HTML for he
 
 This works for RichCommands as well as normal click Commands.
 
+## Render help as JSON
+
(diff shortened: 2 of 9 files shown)


dwreeves:
FWIW, you could do `--json` if you wanted, the `allow_interspersed_args=False` thing makes any options you add before the `script_and_args` argument safe to add, so there is no technical limitation, it's just an API design choice. And maybe that's the correct API design choice? I don't know!!! 😅 I didn't consider it when I proposed adding this. I see the argument for it being that `--output` is a decision for rendering, whereas `--json` fundamentally changes the help text path. So it could be the correct API decision. What do you think?

ewels:
I guess one benefit of `--json` would be that this example gets simpler:

```bash
rich-click -c '{"help_json": true}' --output=svg [cmd] --help-json
```

instead becoming this?

```bash
rich-click --json --output=svg [cmd] --help
```

I'm not sure I have a strong preference either way. Honestly I can't see many people saving SVGs of JSON output, so combination of the two probably isn't a common scenario 😅

ewels:
Closing in favour of #337

PR #336, opened 2026-06-21
Add native asyncclick support via patch(module=asyncclick)

### LINKED ISSUE #288: Consider supporting asyncclick
Description: Rich-click currently doesn't work with [asyncclick](https://github.com/python-trio/asyncclick) due to differences in class inheritance and method resolution order (MRO). When using rich-click with asyncclick commands, we encounter exceptions because rich-click's formatting logic doesn't recognize asyncclick's command classes, and we have to do some awkward monkeypatching to get rich formatted help pages working with an async click backend.

Root Cause:

* Rich-click expects standard Click `Command` and `Group` classes.
* AsyncClick provides `AsyncCommand` and `AsyncGroup` classes that inherit from Click but have different MRO
* Rich-click's detection/formatting logic doesn't recognize asyncclick command types
* This causes rich-click to not apply its enhanced formatting to async commands

What We're Asking For: Official asyncclick compatibility by:

* Recognizing asyncclick command and group types in rich-click's formatting logic
* Ensuring rich-click's `rich_format_help()` works with asyncclick's class hierarchy
* Testing compatibility with asyncclick's async command execution

Why This Matters: AsyncClick enables native async/await CLI commands which are essential for applications that need:

* Concurrent operations and async I/O
* Integration with async frameworks (aiohttp, asyncpg, etc.)
* Proper async context management
* Plugin architectures with async lifecycle hooks

We considered wrapping commands in an async context using anyio or trio, but this has drawbacks when it comes to error and context propagation. What we are currnently doing is using a custom `AsyncRichCommand` and `AsyncRichGroup` inheriting from the AsyncClick classes, but overriding format_help to call rich_format_help. Though, this works, it's not a good long term solution and I have had to pin to an older version of rich-click because newer versions seem to be incompatible with this approach.

gabloe:
Fixes #288 

[asyncclick](https://github.com/python-trio/asyncclick) is a fork of Click whose
 command machinery (`main`/`invoke`/`make_context`) is asynchronous, letting command
 callbacks be coroutines. Previously rich-click could not format an asyncclick CLI:
 its formatting was bound to concrete Click-based classes, and detection used
 `isinstance` checks against `click.*` types that asyncclick's parallel class tree
 does not subclass.

 This PR makes rich-click format asyncclick CLIs the same way it formats Click CLIs.
 End users call `patch(module=asyncclick)` once, then write a completely vanilla
 asyncclick CLI and get Rich help panels and Rich error panels with no rich-click
 specific code. asyncclick stays an **optional** dependency
 (`pip install rich-click[async]`); rich-click's core never imports it.

 ```python
 import asyncclick
 from rich_click.patch import patch

 patch(module=asyncclick)

 @asyncclick.group()
 async def cli(): ...

 @cli.command()
 @asyncclick.argument("name")
 async def greet(name):
     asyncclick.echo(f"Hello {name}")
 ```

 The change is structured so the synchronous Click path is behaviorally unchanged:
 concrete classes keep identical names and MRO, and no existing test output changes.

 ### Changes by file

 **`src/rich_click/_click_types_cache.py`** -- Fork-agnostic detection. Adds
 `register_click_impl(module)` (idempotent) plus `is_argument`/`is_command`/`is_group`/
 `is_option`/`is_parameter` `TypeIs` helpers and type tuples. This replaces hard-coded
 `isinstance(x, click.X)` checks with a registry that any Click-compatible fork can opt
 into, so the renderer recognizes asyncclick's `Argument`/`Option`/`Group`/etc.

 **`src/rich_click/rich_context.py`** -- Splits a baseless `RichContextMixin` (all Rich
 state and formatting) from the concrete `RichContext`. Under `TYPE_CHECKING` the mixin
 uses `click.Context` as a base so static analysis still sees inherited attributes; at
 runtime it is baseless (`object`) so it can compose with either Click's or a fork's
 `Context`.

 **`src/rich_click/rich_command.py`** -- Splits baseless `RichCommandMixin`/
 `RichGroupMixin` from the concrete `RichCommand`/`RichGroup`, which keep identical names
 and MRO. Moves `_error_formatter` into `RichCommandMixin` (it is formatting, shared by
 sync and async) and makes its context check `isinstance(ctx, RichContextMixin)` so it is
 fork-agnostic. Execution-flow methods (`main`, `get_help_option`, `to_info_dict`) stay on
 the concrete classes so they are not shared into the async subclasses.

 **`src/rich_click/rich_help_rendering.py`** and **`src/rich_click/rich_panel.py`** --
 Replace the in-line `isinstance(..., click.X)` checks with the new `is_*` detection
 helpers so help and panel rendering work on asyncclick's types as well as Click's. No
 rendering behavior changes for Click.

 **`src/rich_click/rich_async_command.py`** *(new)* -- The asyncclick bridge. Defines
 `RichAsyncContext`/`RichAsyncCommand`/`RichAsyncGroup`/`RichAsyncCommandCollection` by
 composing the baseless mixins onto asyncclick's async bases. Provides an async `main`
 that mirrors asyncclick's async `main` but routes `ClickException`/`Abort` through
 rich-click's formatter (so errors render as Rich panels), and an async `to_info_dict`
 (asyncclick's is a coroutine). Imports asyncclick at module level (this module is only
 imported when the optional dependency is present) and calls
 `register_click_impl(asyncclick)` on import so the classes also work when used directly,
 not only via `patch()`.

 **`src/rich_click/patch.py`** -- Adds a `module=` parameter to `patch()` and a
 `_patch_async_module` helper. When a module is passed, `patch()` dispatches on whether
 `module.Command.main` is a coroutine (is this actually an async fork) rather than on the
 module name; async forks get their `Command`/`Group`/`CommandCollection` and `command`/
 `group` decorators swapped for the `RichAsync*` versions, mirroring how the existing
 Click path swaps `click.Command` for `RichCommand`. The synchronous Click `patch()` path
 is unchanged.

 **`pyproject.toml`** -- Adds the `optional-dependencies.async = ["asyncclick>=8.1"]`
 extra (`rich-click[async]`); a version-marked `asyncclick` in the `dev` extra so the test
 suite exercises it on Python >= 3.10 and skips cleanly on older versions; a mypy override
 for `asyncclick.*` (`ignore_missing_imports`); and a narrowly-scoped mypy override for
 `rich_click.rich_async_command` disabling only `misc` and `attr-defined`. Those two codes
 are unavoidable structural artifacts of composing rich-click's Click-typed mixins with
 asyncclick's separately-typed parallel class tree (diamond MRO); the bug-catching codes
 (`override`, `index`, `return-value`, `arg-type`) stay enabled and caught a real async/
 sync `to_info_dict` mismatch during development.

 **`examples/13_async.py`** *(new)* -- A runnable example: a vanilla asyncclick CLI
 rendered via `patch(module=asyncclick)`. It guards the asyncclick import and exits 0 when
 the optional dependency is absent, so the existing `python examples/*.py --help` CI
 passes on a base install.

 **`tests/test_asyncclick.py`** *(new)* -- Nine tests, gated by `importorskip`, covering
 both entry points: the `RichAsync*` classes used directly (type detection, Rich help
 panels, argument/option classification, async execution with shared `ctx.obj`, Rich error
 panels) and `patch(module=asyncclick)` end to end (run in a subprocess, since `patch()`
 mutates the module globally and is irreversible).

 ### Compatibility and risk

 - asyncclick is optional; nothing in rich-click's import path requires it.
 - Concrete Click classes keep their names and MRO; the synchronous `patch()` path is
   untouched.
 - Full suite: 152 passed, 6 skipped. mypy at the existing baseline (0 new errors). ruff
   clean.
 - On interpreters without asyncclick, the example exits 0 and the asyncclick tests skip.

 ### Test plan

 ```bash
 pip install -e ".[dev]"
 pytest tests/
 python examples/13_async.py --help
 python examples/13_async.py greet World
 python examples/13_async.py greet     # Rich error panel, exit 2
 ```


gabloe:
@dwreeves Please let me know if this change is something you would accept. I tried to make it as minimally invasive as I could. I am happy to make changes if there are any concerns. Thanks!

dwreeves:
Hi @gabloe, thanks for the PR. I'll take a look when I have a moment (probably won't be for at least 2 weeks, earliest possible for me would be July 5th). I also appreciate that you attempted to make it minimally invasive; that's definitely in the spirit of the acceptance criteria for something like this.

I don't oppose asyncclick support on principle, but it's not something I want to spend a lot of time maintaining. Honestly I still frankly don't fully understand the use case of asyncclick and I'm personally a fan of decoupling interfaces from core code and owning my own event loops. Although, I know some people like asyncclick, which is good enough of a reason to support it **so long as it doesn't create a maintenance burden.** So what I'll really be looking for in this PR (again, when I get to have a closer look) is whether the abstractions here are sufficiently "hands-off." The less asyncclick "plumbing" that we need to port for whatever reason, the better. I do wonder how doable this is in practice, although I cannot say I'm familiar with how asyncclick works under the hood.

---

We don't have any policy around AI generated contributions and we do not oppose AI generated contributions (won't speak on behalf of Phil but I believe that's his position), but just for my own information and for my own accounting, can I get a rough estimate of how much of this is AI generated?

On that note I'll ask Phil if we should have an AI disclosure in our contributing guidelines (currently our contributing guidelines are from the time before AI coding was good and omnipresent); I think that would be reasonable.

gabloe:
Hi @dwreeves Thanks! I totally understand timing issues, July works.

On the maintenance concern, this was the central design goal. I did not want to introduce any asyncclick specifics that you or the other maintainers would need to take forward. No asyncclick execution logic is reimplemented. The only async-specific code is a thin bridge ( rich_async_command.py ) whose async methods just await super() and hand off to rich-click's existing logic. There should not be any ported plumbing to keep in sync with asyncclick internals. To support that, I had to implement helper functions to enable registering alternate click implementations ( register_click_impl() ) since rich-click originally used isinstance guards that referenced click-specific classes, which asyncclick does not use. Those helpers are fork-agnostic by design: they let *any* Click-compatible implementation register itself, so there is nothing asyncclick-specific baked into core rich-click.

On the AI question, happy to be transparent: I wrote the initial changes to support asyncclick based on a POC CLI that I built which uses rich-click + asyncclick (monkeypatch approach). I verified those changes with AI and then used AI to make sure that I covered all of the edge cases with documentation/comments. I would say this is roughly 50% implemented by AI.

gabloe:
Hi @dwreeves, just curious if you had a chance to review this yet?

dwreeves:
Hi @gabloe I'm looking at it today

dwreeves:
Sorry, I spent my time today fixing issues in rich-click's CI. Soon!

dwreeves:
Thanks for fixing the merge conflicts with the main branch!

gabloe:
@dwreeves No problem. Looks like there are more conflicts :'( Fixing those now

gabloe:
Hi @dwreeves, any update on the review?

ewels:
Hi @gabloe,

Apologies for the wait. Whilst I'd generally consider @dwreeves as better than me at reviewing this type of PR, I thought I'd try to step in with a review to get this moving. I've throw my various Claude-assisted review processes at the diff and come up with the following review points. I've taken what it wrote and rewritten it a bit to try to make it easier to follow.

----

Code run against click 8.4.2 and asyncclick 8.4.2.1 and 8.1.8 on Python 3.13. Reproduction steps are below.

Short version: the design is right, but there are three blockers. All three have small fixes.

## The design

Splitting the formatting behaviour into baseless mixins and leaving the concrete classes as mixin-plus-click is the correct way to do this. It means rich-click's core never imports asyncclick, the sync classes keep their exact names and MRO, and the formatting code is shared rather than forked. The `TYPE_CHECKING` trick (click bases for the type checker, `object` at runtime) is a neat way to keep static typing on a class that can't have a real base.

`_make_param_metavar` is a bug fix worth keeping even if the rest of this PR were dropped. Deciding how to call `make_metavar` by checking the *installed click version* is wrong. For example, if asyncclick 8.4 parameter renders in a process with click 8.0 installed, because the version flag describes the wrong library. Inspecting the actual signature fixes a real latent bug on the sync side too.

I also checked that the sync path is genuinely untouched: 160 tests pass, no snapshot churn, and mypy reports the same 45 errors in 2 files on both `main` and this branch. A security-focused pass over the diff found nothing.

## 1. The declared asyncclick floor crashes

`pyproject.toml` declares `asyncclick>=8.1`. But `rich_async_command.py` calls `await ctx.aexit()` and reads `asyncclick.exceptions.NoArgsIsHelpError`, and neither exists before 8.2.2. PyPI has 8.1.3.1 through 8.1.8 sitting inside that range.

Installed 8.1.8 in a clean venv against this branch. Both paths die immediately:

```python
# any successful run
  File "rich_click/rich_async_command.py", line 88, in main
    await ctx.aexit()
AttributeError: 'RichAsyncContext' object has no attribute 'aexit'

# any CLI error
  File "rich_click/rich_async_command.py", line 93, in main
    if isinstance(e, asyncclick.exceptions.NoArgsIsHelpError):
AttributeError: module 'asyncclick.exceptions' has no attribute 'NoArgsIsHelpError'
```

So `pip install rich-click[async]` can resolve to a combination where every invocation is a traceback, not a degraded render. The sync path guards the same click API behind `CLICK_IS_BEFORE_VERSION_82`; the async copy dropped the guard.

Fix: floor of `8.2.2`, plus one set of CI tests pinned to it. The CI matrix currently pins `click` across five versions but lets `asyncclick` float to latest, so nothing tests the minimum.

## 2. The async patch wires decorators differently from the sync one

This one causes two separate symptoms, which is why it's worth looking at the cause first.

`patch()` on click replaces `click.command` with `rich_click.decorators.command`. `_patch_async_module` instead keeps asyncclick's own decorator and only sets `cls` on it:

```python
def _command(*args, **kwargs):
    kwargs.setdefault("cls", RichAsyncCommand)
    return _orig_command(*args, **kwargs)
```

First symptom — the bare decorator form stops working. asyncclick's decorator asserts `cls is None` when applied without parentheses, and now `cls` is always set:

```python
patch(module=asyncclick)

@asyncclick.command          # fine before patch() was called
async def c(): ...
# AssertionError: Use 'command(cls=cls)(callable)' to specify a class.
```

Bare `@asyncclick.group` happens to survive, which makes it look arbitrary from the outside. Neither the example nor the tests use the bare form.

Second symptom — `rich_config()`, `@option_panel` and `@command_panel` silently do nothing. Those decorators stash their settings on the function object as `__rich_context_settings__` and `__rich_panels__`, and rich-click's decorators drain them on the way past. asyncclick's don't know they exist, so the settings sit on the callback forever. Same CLI, same decorator, two outcomes:

```
patch() on click                     patch(module=asyncclick)
╭─ MY OPTS ──────────────╮           ╭─ Options ──────────────╮
│ --help  Show this ...  │           │ --help  Show this ...  │
╰────────────────────────╯           ╰────────────────────────╯
```

Both from `@rich_config(help_config=RichHelpConfiguration(options_panel_title="MY OPTS"))`. No warning — the title is just ignored.

Useful diagnostic: `rich_config` *does* work through the other documented entry point, `rich_click.group(cls=RichAsyncGroup)`. That rules out the mixins and points at the routing.

Having `_command` and `_group` call `rich_click.decorators.command(..., cls=RichAsyncCommand)` fixes both symptoms, and it's already the path `@cli.command()` takes.

## 3. The mixin split stops three-quarters of the way

The `isinstance(x, click.X)` checks were all converted to the new `is_*` helpers. The checks against rich-click's *own* concrete classes were not, and `RichAsyncCommand` is not a subclass of `RichCommand`:

```python
if cls and not issubclass(cls, RichCommand):                  # rich_command.py:454, :509
elif typ == "commands" and isinstance(command, RichGroup):    # rich_panel.py:660
```

The first pair pops `panel` and `aliases` out of kwargs so they never reach the command object. The third skips the `_panel_command_mapping` fallback. Result, from `@cli.command(panel="Custom Panel", aliases=["g"])`:

```
sync                                  async
╭─ Custom Panel ─────────╮            ╭─ Commands ─────────────╮
│ greet   g   Greet.     │            │ greet       Greet.     │
╰────────────────────────╯            ╰────────────────────────╯
```

Dispatch by alias still works; only the display drops it. There's a fourth at `rich_panel.py:380` that's benign today but will bite a `RichAsyncCommand` subclass overriding `get_rich_table_row`.

`help_to_stderr` is the same story one level up. `get_help_option` was deliberately held on the concrete class as an "execution-flow method", but it isn't execution flow — it's what installs rich-click's `show_help` callback, which honours `help_to_stderr`, uses `print()` over `click.echo()` so `force_terminal` survives, and sets `cls=RichOption`. On async, help goes to stdout where sync sends it to stderr.

Widening these to the mixins is worth doing on its own merits. They're the kind of concrete-class assumption that will get in the way when command groups and aliases become panels in 2.0. A grep turns up six such sites; three matter today.

## Why CI didn't catch any of this

The tests assert on substrings — `BOX_BORDER in result.output`, `"Options" in result.output`. A substring check passes whether or not the right panel was drawn, so all nine tests are green while `panel=`, `aliases=` and `rich_config` do nothing. Every module in `tests/help/` uses `inline_snapshot` against full rendered output, and the autouse fixture in `conftest.py` already pins `WIDTH`, `COLOR_SYSTEM` and `FORCE_TERMINAL`, so asyncclick output is just as deterministic. Converting the three help tests would have caught all of section 3.

Coverage of `rich_async_command.py` is 56%. Nothing exercises `RichAsyncCommandCollection`, `standalone_mode=False`, `Abort` or `to_info_dict`, which is most of section 1.

And the examples job runs with `--no-default-groups`, so asyncclick isn't installed, `examples/13_async.py` prints its "requires asyncclick" message and exits 0. That job proves nothing about the example.

Of these, the snapshot tests are the highest-leverage change in the whole review. They're what turns a future asyncclick release into a failing test instead of silent formatting drift.

## On the maintenance question

@dwreeves asked specifically how much asyncclick plumbing this drags in, here's an analysis of that:

The class architecture is about as hands-off as it gets: mixins, an opt-in registration hook, an optional dependency, no asyncclick imports in the core. That part of the PR's claim holds.

`RichAsyncCommand.main` does not. It's a near-verbatim copy of asyncclick's `main` with two rendering branches swapped, resting on five private or version-sensitive APIs: `_expand_args`, `_detect_program_name`, `_main_shell_completion`, `PacifyFlushWrapper` and `aexit`. That's roughly 60 lines to re-sync whenever asyncclick's `main` changes, and section 1 is that coupling already biting before the feature has shipped.

The sync `RichCommand.main` is itself a fork of click's and carries a comment explaining why, so this is parity rather than something new. But it's worth one iteration checking whether the two rendering branches can be reached by overriding only the exception display. If they can't, merging it is defensible — the PR description just shouldn't say there's no ported plumbing.

## Missing pieces

`rich_command.pyi` wasn't updated. The package ships `py.typed` and the stub shadows the `.py`, so the new mixins — the extension seam this PR advertises — don't exist for type checkers:

```
error: Module "rich_click.rich_command" has no attribute "RichCommandMixin"  [attr-defined]
```

The PR works around this by disabling `attr-defined` for the whole `rich_async_command` module, which also suppresses genuine typos in the newest code in the diff.

No docs page, no nav entry, no changelog bullet. `docs/documentation/typer_support.md` is the template for exactly this kind of optional integration. The page should also state what a user can't discover from the code: the patch is global and irreversible, and which decorator features are supported.

The PR description is out of date in two places — it says the mypy override disables only `misc` and `attr-defined` (it disables five codes), and that there's no ported asyncclick plumbing.

## Smaller items

| Where | What |
|---|---|
| `patch.py` | `patch(module=X)` where X is synchronous, lacks `Command`, or is *a string* silently patches click instead and returns as though it worked. The `TypeError` meant to catch this is unreachable — `patch()` already ran the same check before dispatching. |
| `rich_async_command.py` | `group_class = RichAsyncGroup` instead of click's `type` sentinel, so `class MyGroup(RichAsyncGroup)` loses its own type on nested groups. `RichGroupMixin.group()` already implements the sentinel; using it also removes the post-class assignment and the comment saying it's required. |
| `patch.py` | `_patch_async_module` sets `OVERRIDES_GUARD = True` — module state governing *click* commands — for a call that only asked to patch asyncclick. It's a no-op for the async tree, so this is cross-talk with no upside. |
| `rich_async_command.py` | `RichAsyncCommandCollection` has no `format_help` override where `RichCommandCollection` does. Harmless only because of the no-op above. |
| `patch.py` | New docstring says `patch_rich_click ... Defaults to False`; the signature default is `True`. |
| `__init__.py` | The `RichAsync*` classes aren't exported, unlike every sync counterpart. `RichAsyncMultiCommand` also adds a name on the 2.0 deprecation path without the warning `RichMultiCommand` carries. |
| `rich_async_command.py` | `print(e.format_message())` where the sync path uses `print(e.message)`. Same result today, gratuitous divergence. |
| `_click_types_cache.py` | `is_option`, `is_command` and `is_parameter` are barely used — `is_option` only in the new test file. |
| `rich_command.py` | `RichGroupMixin.command()` builds asyncclick command objects using *click's* `command()` decorator. It works because both forks share the `__click_params__` protocol, but that coupling has no comment marking it. |
| `tests/` | Globals are restored to hard-coded `False`/`True` rather than saved values — `monkeypatch.setattr` is safer. Subprocess tests set width 80 where `conftest` uses 100. |
| `pyproject.toml` | The `; python_version>='3.10'` marker on the dev dep is redundant now the project floor is 3.10. |

## Suggested order

1. Raise the asyncclick floor to 8.2.2 and add a CI leg pinned to it.
2. Route the async decorator wrappers through `rich_click.decorators`, and widen the four concrete-class checks to the mixins. Around ten lines between them, and they close the parity claim.
3. Convert the help tests to `inline_snapshot` so that parity stays closed.
4. Stub, docs page, changelog bullet, and the two stale claims in the description.

Things I checked that are fine: error-config inheritance through nested async groups, `CommandCollection` rendering, `no_args_is_help`, sync MRO equivalence, config-driven `COMMAND_GROUPS`/`OPTION_GROUPS` panels, and the example's use of `USE_RICH_MARKUP`.

ewels (changes requested):
Found some issues that need resolving before merge. Please see detailed review in PR comment.

PR #339, opened 2026-07-22
Fix ctx.forward() to rich commands failing when patching

### LINKED ISSUE #338: `ctx.forward(rich_cmd)` fails when patching

JSap0914:
## Summary

Fixes #338.

After `rich_click.patch.patch()`, calling `ctx.forward(rich_cmd)` raised:

```
TypeError: Callback is not a command.
```

### Root cause

`patch()` rebinds `click.Command` / `click.core.Command` to a rich-click
subclass (`_PatchedRichCommand`). Click's own `Context.forward` /
`Context.invoke` resolve the bare name `Command` from `click.core` at call
time and use it in an `isinstance` guard:

```python
# click/core.py
def forward(__self, __cmd, *args, **kwargs):
    if not isinstance(__cmd, Command):   # Command == _PatchedRichCommand after patch()
        raise TypeError("Callback is not a command.")
```

Because the patched class is a *subclass*, the guard is narrowed: any genuine
Click command that is not that exact subclass fails the check. This affects
`RichCommand` instances built via `rich_click.command()`, plain `click.Command`
objects created before `patch()` ran, and third-party commands. `invoke()` has
the same narrowed check and would otherwise mis-dispatch the command as a raw
callback (`cmd(**params)`).

### Fix

Override `forward` / `invoke` on `RichContext` to validate against the genuine
`click.Command` base class, recovered by walking `click.Command.__mro__` so it
still works while `click.Command` is patched. When unpatched, `click.Command`
*is* the genuine base, so behavior is identical (both overrides delegate to
`super()`), and `super().invoke()` still handles all normal command dispatch.

### Tests

Added `tests/test_patch.py` with subprocess-based regression tests (patching
mutates global Click state), covering:

- `ctx.forward()` to a `rich_click.command()` command under `patch()`
- `ctx.forward()` to a base `click.Command` created before `patch()` ran

Both fail on `main` with the exact `TypeError` from the issue and pass with the
fix. `ruff check`, `ruff format`, and `mypy` are clean on the changed files.

### DIFF
--- a/src/rich_click/rich_context.py
+++ b/src/rich_click/rich_context.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, Any, Literal, Mapping, Optional, Type, Union
 
 import click
+from click.core import augment_usage_errors
 from click.globals import get_current_context as click_get_current_context
 
 from rich_click.rich_help_configuration import RichHelpConfiguration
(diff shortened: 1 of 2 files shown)


dwreeves:
Not interested in an AI PR for this.

PR #340, opened 2026-08-09
Misc. dev cleanup

dwreeves:
### Fix mypy breakage

Latest version of mypy broke some of our type checks in the pre-commit. I made the couple minor changes to address that. I also fixed an issue in `pyproject.toml` that made it more likely the CI and local were out of sync.

### pyproject.toml: Swap deps to groups

We were using "optional dependencies" for dev and docs, but what we should be using is groups.

The main difference is optional deps are saved as package metadata; groups are not and are solely for the local environment. Groups are the correct way to do this.

### Swap to prek over pre-commit

I've had my eye on prek for a bit. https://github.com/j178/prek It's become quite popular, and I've been using at work the past 6 or so months. It's a little better than pre-commit in my experience.

### Swap to `ruff format` + remove Black because of ruff conflicts

Once I ran `prek update --freeze`, black and ruff started fighting with each other. To fix, I removed Black and added `ruff format`.

All the other changes are the result of `ruff format`.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -21,7 +21,7 @@ jobs:
           activate-environment: true
       - name: Install dependencies
         run:  |
-          uv pip install ".[dev]"
+          uv pip install --group dev .
           uv pip install "mypy<1.16"  # later versions don't work with python 3.8
       - name: Run pre-commit
         run: pre-commit run -a
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -20,7 +20,6 @@ jobs:
           python-version: 3
           activate-environment: true
       - name: Install dependencies
-        run:  |
-          uv pip install --group dev .
+        run: uv sync --active
       - name: Run prek
         run: prek run -a
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -22,6 +22,5 @@ jobs:
       - name: Install dependencies
         run:  |
           uv pip install --group dev .
-          uv pip install "mypy<1.16"  # later versions don't work with python 3.8
-      - name: Run pre-commit
-        run: pre-commit run -a
+      - name: Run prek
+        run: prek run -a
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,7 +14,7 @@ Requirements:
 uv python pin 3.13
 uv venv .venv
 source .venv/bin/activate
-uv sync --extra dev
+uv sync --all-groups
 pre-commit install
 ```
 
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -28,7 +28,7 @@ jobs:
           activate-environment: true
 
       - name: Install dependencies
-        run: uv pip install ".[docs]"
+        run: uv pip install --group docs .
 
       - name: Generate terminal images with rich-codex
         if: github.repository_owner == 'ewels' || github.event_name == 'workflow_dispatch'
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -28,7 +28,7 @@ jobs:
           activate-environment: true
 
       - name: Install dependencies
-        run: uv pip install --group docs .
+        run: uv sync --active --group docs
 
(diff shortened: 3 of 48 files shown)


.github/workflows/pre-commit.yml
ewels, line 1: Rename to `prek.yml` ?

ewels (changes requested):
Great! I've also had my eye on exactly these changes. I've been using the same toolchain for some time as well now (could potentially also use `ty` instead of mypy now as well, but that'd likely be a bigger change and `ty` is still quite new.

One requested change: I've found that most agents are much more familiar with pre-commit than prek, and can cause issues by running the wrong tool. The fix is simple: use a `prek.toml` config file instead of `pre-commit.yml` and generally avoid using the name "pre-commit" anywhere. Then agents are usually fine with it.

CONTRIBUTING.md
ewels, line 21: This section needs updating now.

ewels (commented):
Maybe worth a sweep of the docs to make sure that all descriptions are updated.

ewels:
Note: we have a few other PRs in flight at the moment (I'm back from holiday in a few days and am hoping to revisit mine with a fresh set of eyes). I suggest we merge this one first though and then update others.

ewels:
Merged and added in my suggested changes in 5bd402f2ff55a6be1bdfcc0fa6144936e3d2a5aa

dwreeves:
@ewels I tried to add `ty`, but there were a lot of things flagged (89 things if I recall correctly). I decided to go down that path would be a bigger project requiring its own PR.

FWIW, many of the things `ty` flagged were things like this. One example, but it represents a _lot_ of examples I saw:

```diff
def rich_format_error(
    self: click.ClickException, formatter: RichHelpFormatter, export_console_as: Literal[None, "html", "svg"] = None
) -> None:
    """
    Print richly formatted click errors.

    Called by custom exception handler to print richly formatted click errors.
    Mimics original click.ClickException.echo() function but with rich formatting.

    Args:
    ----
        self (click.ClickException): Click exception to format.
        formatter: formatter object.
        export_console_as: If set, outputs error message as HTML or SVG.

    """
    config = formatter.config
    # Print usage
    if getattr(self, "ctx", None) is not None:
        if TYPE_CHECKING:  # pragma: no cover
            assert hasattr(self, "ctx")
        self.ctx.command.format_usage(self.ctx, formatter)
    if config.errors_suggestion:
        formatter.write(
            Padding(
                config.errors_suggestion,
                config.padding_errors_suggestion,
            ),
            style=config.style_errors_suggestion,
        )
    elif (
        config.errors_suggestion is None
+        and getattr(self, "ctx", None) is not None
        and self.ctx.command.get_help_option(self.ctx) is not None  # type: ignore[attr-defined]
    ):
-        cmd_path = self.ctx.command_path  # type: ignore[attr-defined]
-        help_option = self.ctx.help_option_names[0]  # type: ignore[attr-defined]
```

Explanation:

`click.ClickException` does not have a `ctx`, but `click.UsageError` does.

Here the `getattr(self, "ctx", None) is not None` is essentially acting as a duck-typed `isinstance(self, click.UsageError)`. IIRC this may have been because of old Click.

What's interesting is `ty` doesn't even flag the `help_option_names` as an attr error, it actually says that `ctx` is the attr error! I tried checking if `hasattr(self, "ctx")` fixed the issue (here ty should treat this type in the `elif` block as a `click.ClickException` with "protocol" of `ctx`), but that did not work.

^ The thing is, a ton of the errors in ty were like this.

So this leaves me wondering:

- Do we need to configure ty?
- Are _we_ doing something wrong? Again, I don't entirely know, because it flagged `ctx` as violating the type checking and not `command_path`...
- Is this something that would work or require less maintenance on our end when `ty` is out of beta?

It is also worth mentioning, our choice to duck-type over explicit `isinstance()` checks is a function of supporting not only multiple Click versions but also simultaneously both click and rich-click types (e.g. a `click.Parameter` and `rich_click.parameters.RichParameter` must both be renderable). So switching off duck-typed checks could be tricky.

ewels:
Ah interesting! Yes definitely not worth it if it causes trouble. From my POV, types are primarily a convenience thing for us in Python really. We do some fairly odd stuff in this library so I'm not super surprised that some libs flag it as errors 😅

Totally fine to stick with mypy for the foreseeable future for me.

PR #343, opened 2026-08-12
Fix docs versioning

dwreeves:
I noticed "1.9" never drops the `prerelease` tag. This code fixes that.

Also I made it easier to debug this in Github Actions; perhaps a little overkill.


ewels (approved):
Good spot 👍🏻 

Claude found a couple of other very minor tweaks in addition - will add them into the merge to main.

PR #344, opened 2026-08-14
Fix ctx.forward(rich_cmd) after patch() (#338)

### LINKED ISSUE #338: `ctx.forward(rich_cmd)` fails when patching

DSeaStar:
## Summary

- `patch()` replaces `click.Command` with `_PatchedRichCommand`, a *subclass* of `RichCommand`. Click's `Context.forward` / `invoke` then do `isinstance(cmd, Command)` against that subclass, so a normal `RichCommand` is rejected with `TypeError: Callback is not a command.`
- Teach the patched types to recognize the original Click/`RichCommand` instances (looked up on the class itself so Typer patches stay narrow).

Fixes #338

## Test plan

- [x] `pytest tests/test_patch.py` — `ctx.forward(rich_cmd)` and `ctx.invoke(rich_cmd)` after `patch()`
- [x] `pytest tests/test_rich_click_cli.py::test_simple_rich_click_cli` — existing CLI patch path, including subclassed `click.Command`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Fixed `ctx.forward(rich_cmd)` raising `TypeError` after `patch()` replaced `click.Command` with a subclass. [[#338](https://github.com/ewels/rich-click/issues/338)]
 - Dropped support for Python 3.8 and 3.9; the minimum supported version is now Python 3.10.
 - Modernised the codebase to Python 3.10+ syntax (PEP 604 unions, built-in generics) and removed version-conditional shims for Python <3.10.
 - Dev and docs requirements moved from optional dependencies to [dependency groups](https://peps.python.org/pep-0735/), so they are no longer published as package metadata. Install them with `uv sync --all-groups`.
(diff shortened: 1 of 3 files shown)


dwreeves:
We already resolved this in #351.

The approach you outlined is better than the other AI generated PR for this (339, not linking), but it is still not the best solution.

The machinery to fix this was already there. https://github.com/ewels/rich-click/pull/351/changes

```diff
def patch(
    rich_config: Optional[RichHelpConfiguration] = None,
    *,
-   patch_rich_click: bool = False,
+   patch_rich_click: bool = True,
    patch_typer: bool = False,
) -> None:
```

I'm assuming this was AI assisted (which is not a problem, that's where software development is, and the AIs are admittedly pretty good now, and we will accept AI PRs as long as the code is good). I appreciate that it appears you directed the AI to ensure the solution was minimally invasive compared to the other one. I don't say this as a knock on you, just a pithy little observation I'm throwing out into the ether, but it is interesting every AI PR missed that there is a solution that is as simple as changing a harmless default from `False` to `True`. Humans aren't fully replaced just yet!

PR #345, opened 2026-08-14
Fix ctx.forward(rich_cmd) TypeError after patch()

### LINKED ISSUE #338: `ctx.forward(rich_cmd)` fails when patching

DSeaStar:
## Summary

Fixes #338.

`patch()` replaces `click.Command` / `click.core.Command` with `_PatchedRichCommand`, a *subclass* of `RichCommand`. A command created with `@rich_click.command()` is a `RichCommand` instance — a *parent* of that subclass — so after patching:

```python
isinstance(rich_cmd, click.Command)  # False
ctx.forward(rich_cmd)                # TypeError: Callback is not a command.
```

Click's `Context.forward` / `Context.invoke` both gate on that `isinstance` check.

This teaches the patched types (via `PatchMeta.__instancecheck__`) to also accept instances of the original `Command` / `Group` / `CommandCollection` classes captured at import time. The override is stored on each patched class and read from `cls.__dict__` so it is not inherited by Typer patches.

## Test plan

- [x] Reproduced on current `main`: `ctx.forward(rich_cmd)` after `patch()` raises `TypeError: Callback is not a command.`
- [x] After this change the same script prints `hello sea` and exits 0
- [x] `pytest tests/test_patch.py` — 4 passed (isinstance for command/group; `forward` and `invoke` after `patch()` in a subprocess)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Fixed `ctx.forward(rich_cmd)` raising `TypeError` after `patch()` replaced `click.Command` with a subclass. [[#338](https://github.com/ewels/rich-click/issues/338)]
 - Dropped support for Python 3.8 and 3.9; the minimum supported version is now Python 3.10.
 - Modernised the codebase to Python 3.10+ syntax (PEP 604 unions, built-in generics) and removed version-conditional shims for Python <3.10.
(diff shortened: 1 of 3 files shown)


DSeaStar:
Closing as a duplicate of #344, which already fixes the same #338 TypeError after patch().

PR #337, opened 2026-06-27
Machine-readable CLI help: v2

ewels:
Ok, this turned into a bit of a beast, sorry.

## Markdown and other formats

I've been mulling over PR #331 and thinking what else I could do to make the output as LLM-friendly as possible. It struck me that whilst the approach was solid, LLMs actually prefer markdown over JSON, or anything else. The JSON output still seems cool though (rich-click knows about the whole CLI area and can document it, so other can build on top of that).

I didn't want to add loads of additional flags (`--help-markdown`, `--help-json` etc) so instead, in this PR I alter the regular `--help` to be able to _optionally_ take a format argument. I really like this approach because:

1. No new flags
2. Normal `--help` behaviour is identical
3. Net-new functionality only

Because of this, I'm fairly happy shipping it as **default-_on_**, unlike the opt-in approach in #331.

## Full formats and Carapace

With the idea of having others build stuff on the JSON and this new flexibility, I figured it could make sense to have two flavours of JSON and markdown: the progressive disclosure one by default, but also a comprehensive one. So there is `json-full` and `markdown-full`. I could imagine folks using this to generate documentation websites for example, and all sorts.

I also did a bit of research to look for any other standards trying to do similar things (describe CLI command structures) and found [Carapace](https://github.com/carapace-sh/carapace-spec). So I added that too.

The final list is:

1. `--help`
1. `--help markdown`
2. `--help markdown-full`
3. `--help json`
4. `--help json-full`
5. `--help carapace`

The only downside of this approach is that `--help` can't capture invalid strings. So if you make a typo it gets silently swallowed and you just get regular `--help` output back.

In theory everything should be extensible, so others can add their own formats, or customise the outputs somehow.

## Command Examples

Whilst thinking about making the help output more LLM-friendly, it also struck me that they _love_ examples. In fact, humans do too! I've been using https://tldr.sh for years on the CLI and that's all built around having a short list of examples for any CLI tool.

There's no native way to handle this, so I added a new syntax to allow tool developers to specify examples for any command. It takes a description and a command.

```python
@pipelines.command(
    "lint", aliases=["l"],
    examples=[
        ("Lint a release-ready pipeline with full checks", "nf-core pipelines lint --dir ./my-pipeline --release"),
        ("Run only specific tests and write results to Markdown", "nf-core pipelines lint -k files_exist -k nextflow_config --markdown lint-results.md"),
[...]

PR #346, opened 2026-08-17
Blog post: help-text for agents

ewels:
See https://github.com/ewels/rich-click/pull/337


dwreeves:
Great stuff. Benchmarking is 100% the right call, otherwise there is a real question of the point of it. Reducing number of turns is especially great.

I did a skim of this a couple days ago. I unfortunately have a busy weekend ahead of me and I'll do a bigger read when I get a chance, but it is your blog so of course it should be fine as-is.

I do see one very minor issue that I would rather email you about than discuss here.

PR #347, opened 2026-08-31
Improve Typer>=0.26 compatibility handling in patch_typer()

### LINKED ISSUE #330: Typer patching no longer works with `typer≥0.26`
This one might be very difficult to resolve...

ewels:
Attempt to fix https://github.com/ewels/rich-click/issues/330

Or at least, not trigger a hard crash anyway. Catch the error and log a warning, then continue without rich click 🥲

<details>

## Summary
Improved the robustness of `patch_typer()` to gracefully handle incompatibilities with Typer>=0.26, which vendors its own fork of Click with conflicting metaclasses. Instead of allowing exceptions to propagate and crash the CLI, the function now catches build failures, warns the user, and leaves Typer unpatched.

## Key Changes
- **Atomic patching**: Build all patched Typer classes (`TyperCommand`, `TyperGroup`, `TyperOption`, `TyperArgument`) inside a single try/except block before assigning any of them. This prevents leaving Typer in a half-patched state if a failure occurs.
- **Graceful fallback**: When patching fails (e.g., due to metaclass conflicts with Typer>=0.26), catch the exception, emit a `RuntimeWarning`, and return without raising. This allows CLIs to continue working with Typer's default help rendering.
- **Improved warning messages**: 
  - Changed the Typer>=0.26 incompatibility warning from an error tone to informational, with a link to documentation
  - Added a new warning for actual patching failures that explains the graceful fallback behavior
- **Documentation**: Added comprehensive "Typer 0.26+ Support" section explaining the metaclass conflict, the fallback behavior, and workarounds.
- **Tests**: Added three test cases covering:
  - No warning when Typer is compatible
  - Warning when Typer version is flagged as incompatible
  - Graceful fallback when metaclass conflicts occur

## Implementation Details
- Introduced `_TYPER_SUPPORT_DOCS_URL` constant for consistent documentation references
- Inverted the version check logic (`if not TYPER_IS_BEFORE_VERSION_026`) to warn about newer incompatible versions
- Store patched classes in local variables before assignment to enable atomic all-or-nothing patching
- Exception handling preserves the original exception type in the warning message for debugging

https://claude.ai/code/session_01Jf1Y8wtMQgogFYgVXXvCcZ

</details>


dwreeves:
That's probably the best approach for now.

---

For what it's worth, on the status of this going forward:

https://github.com/fastapi/typer/discussions/1797#discussioncomment-17153391

> With respect to the vendoring, the main goal was really to be able to have a more independent code base that will allow us to implement novel features more easily. In the coming months, we'll be making some decions with respect to "legacy" Click functionality. Things we'll want to keep, we'll want to document more clearly in the Typer docs & public API, and we'll reimplement such functionality properly within Typer's own code base. Things we won't want to maintain will probably be removed from the vendored code base.

^ In other words, their click vendoring is not really stable and it appears like `patch_typer()` is not long for this world. Oh well. I have a fondness for the incredibly cursed metaclass Python code I wrote to make that work... 🥲 But it is what it is.

They are potentially interested in implementing themes, so Typer users may one day be able to experience the joy of having access to different color schemes.

---

Tangentially, a funny thing I found out, possibly something I inspired Michael Waskom to do (? unclear, but I did interview at Modal in January), is that Modal moved off Typer and implemented their own minimal "rich click". `uvx modal --help`. https://github.com/modal-labs/modal-client/blob/main/py/modal/cli/_help.py Instead of doing it in Typer. I also quite like the theme tbh, it's like `modern` theme but a lot slimmer.

ewels:
Hah, fun! We should backport their theme here 😅

dwreeves:
I agree, it's pretty nice! Wish I had come up with it.

The one issue is the Apache License being parasitic. Might want to do a "clean room" version, so no referencing of source code, just to be safe.

dwreeves (approved):

PR #351, opened 2026-09-05
Fix patching issue with ctx.forward()

### LINKED ISSUE #338: `ctx.forward(rich_cmd)` fails when patching

dwreeves:
Resolves #338.


dwreeves:
Implemented in #352

PR #352, opened 2026-09-05
1.9.9 release

dwreeves:
Did some git cherrypicking as I am not a fan of removing Python versions on patch updates, even if they are EoL.


dwreeves:
@ewels I'm going to release this off `1.9.x` and then reverse-merge this into main. So the `1.9.9` release head will point to commit `c962bc0`, not main.

I ended up forking off the commit right before the Python 3.9/3.10 removal.

The objective is to fix click 8.5 compat + fix a (minor, obscure) bug with minimal other disruptions. The reverse-merge should keep the commit history relatively clean while allowing for `main` to be in a slightly less stable state.

ewels:
@dwreeves sounds great, thank you!

Issue #353, opened 2026-09-05
Should `rich_click/__init__.py` rely on `__getattr__` and wrap all imports in `TYPE_CHECKING`?


dwreeves:
The old pattern used to be a `from click import *`. The current pattern is something like this:

```python
__version__ = "1.9.9"

# Import the entire click API here.
# We need to manually import these instead of `from click import *` to force
# mypy to recognize a few type annotation overrides for the rich_click decorators.
from click.core import Argument as Argument
from click.core import Command as Command
from click.core import CommandCollection as CommandCollection
from click.core import Context as Context


# ... etc.
# ...
# and here are the rich-click imports

from rich_click.decorators import argument as argument
from rich_click.decorators import command as command
from rich_click.decorators import command_panel as command_panel
from rich_click.decorators import confirmation_option as confirmation_option

# ... etc.
# ...

def __getattr__(name: str) -> object:
    from rich_click._compat_click import CLICK_IS_BEFORE_VERSION_9X

    if name == "RichMultiCommand" and CLICK_IS_BEFORE_VERSION_9X:
        import warnings

        warnings.warn(
            "'RichMultiCommand' is deprecated and will be removed in Click 9.0. Use 'RichGroup' instead.",
            DeprecationWarning,
            stacklevel=2,
        )
        from rich_click.rich_command import RichMultiCommand

        return RichMultiCommand

    else:
        import click

        return getattr(click, name)
```

We initially implemented the current `rich_click/__init__.py` pattern to assist in type-checking, as per the comment in the file. The old pattern did not type-check in a friendly manner.

This was fine for a bit. However, Click and other Pallets projects repos have been a little more aggressive in later versions about moving things around and deprecating things, and this caused an issue breaking CI for Litestar, which asserts no warnings: https://github.com/litestar-org/litestar/issues/5020. And certainly caused some minor annoyances for some other users as well.

It may therefore be more appropriate to do something like this:

```python
import typing as _t

from click.core import Argument as Argument
from click.core import Command as Command
from click.core import Context as Context
from click.core import Group as Group
from click.core import Option as Option
from click.core import Parameter as Parameter

if _t.TYPE_CHECKING:
    # These are secondary in the core Click API; move here to be safe.
    from click.core import CommandCollection as CommandCollection
    from click.decorators import make_pass_decorator as make_pass_decorator
    from click.decorators import pass_obj as pass_obj
    from click.exceptions import Abort as Abort

[...]

PR #355, opened 2026-09-06
Resolve the secondary Click re-exports lazily via `__getattr__`

dchaudhari7177:
Implements the `__getattr__` + `TYPE_CHECKING` pattern from #353.

**Split follows the proposal exactly.** The six primary `click.core` names — `Argument`, `Command`, `Context`, `Group`, `Option`, `Parameter` — stay eager; the other 44 move under `TYPE_CHECKING` and are served by the `__getattr__` fallthrough to `click` that already existed. `CommandCollection` moves with the secondary set, as the issue suggested.

I initially moved *all fifty*, and that was wrong for a reason worth recording: it broke `test_no_rich_imports.py::test_imports_during_execution`. That test does `importlib.reload(rich_click)` and then asserts `any(m.startswith("click."))` as a control — a canary that the import recorder captured anything at all, without which its three negative assertions (`not any rich`, `not any markdown_it`, `not any importlib`) would pass vacuously. With no eager `from click...` line left, the reload recorded no `click.*` import and the canary died. Keeping the primary six eager preserves it, so the guard stays honest and no test needed changing.

**Verified:**

```
$ python -W error::DeprecationWarning -c "import rich_click"
(clean)

$ python -W error::DeprecationWarning -c "import rich_click; rich_click.RichMultiCommand"
DeprecationWarning: 'RichMultiCommand' is deprecated ...
```

So the warning is not swallowed — it moves to the code that actually uses the name, which is the point.

All 50 re-exports still resolve to the *identical* Click objects (`getattr(rich_click, n) is getattr(click, n)` for every one). `Command` is in `__dict__`; `Choice` is not, before or after access.

**`__dir__` added**, as you suggested. Module `__dir__` defaults to `__dict__` and the lazy names are not in it, so without this they'd disappear from tab-completion and `help()`. The name list is written out explicitly rather than derived from `dir(click)`, so the public surface stays a decision made in this file rather than whatever Click version happens to be installed.

**Test suite:** I diffed the full failure set against the unmodified tree — **200 failures/errors before, 200 after, zero regressed and zero newly failing**. (The suite is largely red on my Windows box for unrelated console-encoding reasons; the diff is the meaningful number, not the absolute.) `mypy src/rich_click` reports the same 47 pre-existing errors as `main`. `ruff check` clean on `src/` and `tests/`.

**New test** — `tests/test_lazy_click_reexports.py`, 104 cases: every declared re-export is the identical Click object, the six primary names are eager, the other 44 are absent from `__dict__`, `dir()` lists them all and is sorted, a fresh interpreter imports clean under `-W error::DeprecationWarning`, and a deprecated name still warns at the use site. It also guards its own parser — an empty name list would make the parametrized cases vacuous.

One thing I did *not* do: cache resolved names into `globals()`. You called the per-access `__getattr__` cost trivial and it is, and caching would make `"Choice" in rich_click.__dict__` depend on access history, which is a worse thing to have to reason about than a dict lookup.

### DIFF
--- a/src/rich_click/__init__.py
+++ b/src/rich_click/__init__.py
@@ -7,62 +7,84 @@ customization required.
 
 from __future__ import annotations
 
+import typing as _t
+
 
 __version__ = "1.9.9"
 
-# Import the entire click API here.
-# We need to manually import these instead of `from click import *` to force
-# mypy to recognize a few type annotation overrides for the rich_click decorators.
+# The primary Click API, imported eagerly.
+# We name these individually rather than `from click import *` to force mypy to
+# recognize a few type annotation overrides for the rich_click decorators.
+#
+# These six are the load-bearing names -- every rich-click subclass derives from
+# one -- and Click is not going to move or deprecate them, so importing them
+# costs nothing and keeps `rich_click.Command` resolvable without a __getattr__
+# hop on the hottest path.
 from click.core import Argument as Argument
 from click.core import Command as Command
-from click.core import CommandCollection as CommandCollection
 from click.core import Context as Context
 from click.core import Group as Group
 from click.core import Option as Option
 from click.core import Parameter as Parameter
-from click.decorators import make_pass_decorator as make_pass_decorator
-from click.decorators import pass_obj as pass_obj
-from click.exceptions import Abort as Abort
-from click.exceptions import BadArgumentUsage as BadArgumentUsage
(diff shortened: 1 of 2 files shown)


dwreeves:
I don't understand the point of these slop PRs. Do you think I can't also take an issue description and plop it into an AI agent?

PR #356, opened 2026-09-06
[Proposal] AI policy

dwreeves:
The AI spammers who are responding to issues I leave open are really annoying.

I would propose the following AI policy. I am specifically targeting people who just spam PRs on open issues.

Right now I just close them, but I would like to be able to say that this is our policy.

Separately, I would also propose that PRs should be tied to issues. I don't believe anyone has gone rogue and just opened up a PR not tied to an issue. But I do believe that proof of work and thought used to be more directly tied to there being code, so it wasn't so bad if PRs weren't tied to issues. Now, that is no longer the case, and there is probably a stronger argument now for issues being tied to PRs.

(I do realize the irony of me proposing something in a PR and not via an issue first 😆)

Let me know what you think.


ewels (approved):
I think this is reasonable 👍🏻 

And I think project maintainers are free to ignore any and all rules as they see fit 😆

PR #357, opened 2026-09-06
Fix deprecation warnings going forward, hopefully

### LINKED ISSUE #353: Should `rich_click/__init__.py` rely on `__getattr__` and wrap all imports in `TYPE_CHECKING`?

dwreeves:
Fixes #353.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Dropped support for Python 3.8 and 3.9; the minimum supported version is now Python 3.10.
 - Modernised the codebase to Python 3.10+ syntax (PEP 604 unions, built-in generics) and removed version-conditional shims for Python <3.10.
 - Fixed docs versioning, so that a released version no longer keeps the `prerelease` alias alongside `latest`. [[#343](https://github.com/ewels/rich-click/pull/343)] ([@dwreeves](https://github.com/dwreeves))
+- Redesigned `__init__.py` to avoid surfacing click's `DeprecationWarning` going forward.
 
 ## Version 1.9.9 (2026-09-05)
 
--- a/src/rich_click/__init__.py
+++ b/src/rich_click/__init__.py
@@ -10,59 +10,66 @@ from __future__ import annotations
 
 __version__ = "1.9.9"
 
-# Import the entire click API here.
+import typing as _t
+
+# Import click API here.
 # We need to manually import these instead of `from click import *` to force
 # mypy to recognize a few type annotation overrides for the rich_click decorators.
 from click.core import Argument as Argument
 from click.core import Command as Command
-from click.core import CommandCollection as CommandCollection
 from click.core import Context as Context
 from click.core import Group as Group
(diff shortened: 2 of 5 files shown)


ewels (approved):
LGTM 👍🏻

PR #358, opened 2026-09-22
Add `square` theme format

ewels:
Adds a new theme format, `square`: identical to the default `box` format, but with square corners instead of rounded ones. Previously the only format with sharp corners was `robo`, which also comes with a lot of extra padding — this gives the sharp corners with the default spacing, usable as `default-square`, `nord-square` and so on without setting any other style options.

The `box` styles are extracted into a module-level `_BOX_STYLES` dict, which `square` spreads and then overrides the three panel box styles, so the two formats can't drift apart.

```
┌─ Options ────────────────────────────────────────────────────────────────────┐
│ *  --name  -n  TEXT  Your name. [required]                                   │
│    --help            Show this message and exit.                             │
└──────────────────────────────────────────────────────────────────────────────┘
```

Also adds a docs tab with a screenshot, a snapshot test, and a changelog entry. Note that `all_themes.svg` (the `rich-click --themes` screenshot) still lists 5 formats — regenerating it locally rewrites all the colours because of a different colour depth, so it needs a run of the rich-codex workflow.


docs/images/code_snippets/themes/themes_ex5.svg
ewels, line 1: Note that this was an old orphaned example, hence it shows as a change rather than a new example.

